PR #19, opened 2015-10-29
Standardize multiple value fields between ui and XML

### LINKED ISSUE #9: Save/Retrieve Form Field Data
Hello,

First - thanks for your work on this utility.

Next set of questions for this:

1. I noted in a previous question asked that you are writing the generated XML to a hidden textarea on screen (cool).  My question is if I want to re-load a previously saved form (e.g. to allow a user to edit a previously created form), do I just set the value of the hidden textarea to the previously saved XML on page load?

2. If I want to add/remove a field type or if I want to add an attribute to a field type, how can I go about doing that?  An example - is for each field, I want the user to be able to select from a dropdown a DB field to map the data to.

Thank you.
Daniel

kevinchappell:
This is a follow-up to resolve #9 as per @dannycamps comment here: https://github.com/kevinchappell/formBuilder/issues/9#issuecomment-152152636


dannycamps:
WOW thanks again.  I hate to keep doing this but now I am getting the field values displayed on screen in place of the labels and the labels are gone from the XML.

For example:
If I create a form with a radio Yes/No column, I am setting the option labels and values respectively to:
Yes / Y
No / N

When I reload the form, it is displaying as:
Y / [blank]
N / [blank]

Original created XML:
`
<field name="radio-group-1446065135152" label="Would you like to join us for a Mel Brooks movie night?" description="" required="false" type="radio-group">
   <option label="Yes">Y</option>
   <option label="No">N</option>
</field>
`

XML displayed after reload:
`
<field name="radio-group-1446065135152" label="Would you like to join us for a Mel Brooks movie night?" style="multiple" description="" required="false" type="radio-group" >
			<option value="">Y</option>
			<option value="">N</option>
		</field>
`

Thanks,
-Daniel

kevinchappell:
@dannycamps Unfortunately you'll need to recreate that field. The XML markup for the options node changed slightly in the last update. Label and values are now stored as `<option value="optionValue">optionLabel</option>`

I updated the [changelog](https://github.com/kevinchappell/formBuilder/blob/master/CHANGELOG.md) to reflect this change.

PR #54, opened 2016-02-07
Bugfix: fields cannot be sorted

kevinchappell:


kevinchappell:
Some background on this bug:
The `disabledFields` allows the development to set fields or arbitrary markup to be prepended or appended to the formBuilder. These fields are not movable and new fields cannot not be placed before the first or after the last field. This PR fixes a bug where users could not order a field to the beginning or end of the formBuilder even when disabledFields were not defined.

PR #62, opened 2016-02-20
Add ability to reinitialize

kevinchappell:
By popular demand, formBuilder can now be reinitialized by calling it again.

```
$('.fb-textarea').formBuilder();
```
screenshot from 2016-02-20 14-24-25

Then later:
```
var opts = {
              messages: {
                text: 'Cool Input'
              }
            };
$('.fb-textarea').formBuilder(opts);
```
screenshot from 2016-02-20 14-24-45

The second call can be made from an ajax or button click callback. With these options, when formBuilder rerenders the default label for `text` input will change from "Text field" to "Cool Input".


wudoz:
I found there are still some items can't be translated.
image

kevinchappell:
@wudoz Fixed in #96

PR #66, opened 2016-02-21
Solved bug of checking checked property of null

basitsattar:
Modified the lines that attaches callbacks to add new options. It was throwing error "trying to get checked property of undefined" due to which it was not possilbe to add new options to radio group.

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1134,11 +1134,16 @@
       _helpers.disabledTT($(this));
     });
 
-    // Attach a callback to add new options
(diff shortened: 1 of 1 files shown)


kevinchappell:
Only issue I found was with the Select field as demonstrated in this video.
[Select field problem](http://www.youtube.com/watch?v=xtTgkjobrxc)

kevinchappell:
I'm going to merge since the remaining issue is minor and can be fixed in the an upcoming round of refactors

PR #86, opened 2016-03-15
package.json fixes

ivuorinen:
Sorry about the extra commits, I thought I was already sending you the pull request.

Currently it's impossible to ``npm install kevinchappell/formBuilder`` because ``Path must be a string. Received [ 'dist/*' ]``.

This patch fixes that, and other validity problems ``package.json`` had.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,14 +1,14 @@
 {
   "name": "formBuilder",
   "version": "1.9.0",
-  "main": [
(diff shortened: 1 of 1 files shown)


kevinchappell:
Excellent! I never did direct install always `git clone` then `npm install` without problem so had no idea. Great fix.

PR #90, opened 2016-03-15
Compatibility issues with IE

kevinchappell:
Added a polyfill file to solve several ES6 issues in Internet Explorer.


domialex:
By the way, **form-render.js** also need the polyfill for assign.

PR #95, opened 2016-03-16
Added dist/form-render.* to files list.

ivuorinen:
Added missing form-render files so they are included with ``npm install kevinchappell/formBuilder --save``.


kevinchappell:
Good call.

PR #106, opened 2016-03-18
update comments

githubjeka:


kevinchappell:
Looks good, only recommendation would be `git config --global core.autocrlf true` to auto-convert line feeds.

PR #117, opened 2016-03-29
Use $ as jQuery.noConflict

mvanvu:
Thanks for awesome FormBuilder
jQuery will conflict "Uncaught TypeError: $ is not a function" when use it in Joomla!.

My simple idea just define a local $ = jQuery.noConflict();

### DIFF
--- a/dist/form-builder.js
+++ b/dist/form-builder.js
@@ -194,7 +194,7 @@ var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol
 
 function formBuilderHelpersFn(opts, formBuilder) {
   'use strict';
-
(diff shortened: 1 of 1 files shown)


kevinchappell:
@mvanvu Thanks for your contribution, will test locally and see if we can add this to 1.9.x.

kevinchappell:
I'm not sure if the issue this PR solved remains as it was never reported. Also the plugin itself should not have a conflict issue since it is wrapped with:
```
(function($) {
  // plugin contents
})(jQuery);
```

Feel free to rebase and resubmit the PR if you feel its still valid but remember edit the `/src` files otherwise changes will be lost on build.

PR #155, opened 2016-05-24
Create fa-IR (Persian/Farsi language)

pooya-eghbali:
Persian/Farsi translation for formBuilder, layout should be RTL. Using formBuilder in my project, this is the least i can do for it.


kevinchappell:
This is fantastic! Thank you!

PR #173, opened 2016-06-21
file with a new language pt-BR(Portugês do Brasil)

AcNeidert:


kevinchappell:
This is excellent, formBuilder has a lot of Portuguese speaking users so this helps a lot. Thank you!

PR #188, opened 2016-07-12
Feature: Number input field

### LINKED ISSUE #185: Number input
After doing some research, it seems this plugin doesn't support input of type="number". Can someone shed some light on that?

kevinchappell:
More browsers are supporting the html5 number input. For this reason the number input is being added to formBuilder but since there are still backward compatibility issues, it is disabled by default.

To enable, remove 'number' the field from the `disabledFields` option.

number-input

Resolves #185


jonrcable:
on point! I literally just started checking this out yesterday replacing an existing lib that doesn't work nearly as well, one of the things missing was the HTML5 number element, this couldn't have been better timing, so high five. One last thing, I wonder if its possible to get a tutorial on how to add additional elements, I was able to modify it enough to remove what I didn't need, like className and roles, I actually just merged my diffs to the numbers release in under a few mins. I will be trying to add a few additional (custom) fields at some point (APIs) in the future, a basic tutorial on adding elements would be very dope! cheers!

kevinchappell:
@jonrcable A tutorial would certainly help in extending formBuilder however most fields not included will have their own special needs and the tutorial may serve as a code tour at best. Unfortunately formBuilder wasn't created with extensibility in mind, [Formeo](https://github.com/Draggable/formeo) however was but is still in early stages of development. For example adding a number element in Formeo would simply be this: 
~~~
{
  tag: 'input',
  attrs: {
    type: 'number',
    min: '',
    max: ''
  }
}
~~~

Regardless I can try to put together a tutorial one of these days. Do you have a field type in mind?

PR #190, opened 2016-07-13
Fast edit options

kevinchappell:
Fields can now be added to the stage by clicking them.
Useful for testing and added many fields quickly.

With sticky controls you can build long forms without needing to scroll back to the top to add the next field.
To enable, set `stickyControls` option to `true`.

Most users will want to edit the field they just added, save them a click by enabling `editOnAdd`. This will automatically close all fields currently being edited and open the field they just added.


Alvareto:
When `editOnAdd` is enabled, only label is shown for each field that is not being edited at the moment - `<input \>` element is not visible.
Also, none of the Header fields are visible.

kevinchappell:
What browser are you seeing this in? I'm not able to replicate the header field issue but the preview is hidden when in edit mode by design. It should be visible for all fields not being edited and is on Firefox and chrome on Debian Linux. I can test later on Windows.

Issue #184, opened 2016-07-07
json implementation


Yc-Chen:
Hi,

I have been looking into integrating formBuilder and a [form generator](http://brutusin.org/json-forms/). If these two are integrated, many features mentioned in other issues will become available or easy to implement. The biggest gap now is between xml and json. I see there are 'coming soon' comment for json in the codes. However what is the progress? I'm willing to implement this if nothing is done yet.

kevinchappell:
@Yc-Chen, that would be a huge help. JSON form data has not been implemented in formBuilder, instead it was added to [Formeo](http://draggable.github.io/formeo/) which is the next major release of formBuilder (explaination of namechange on formbuilder.online News section). However Formeo is not ready for production and a number of people will likely stick with formBuilder even after it is so having JSON as an option would be huge.

blaketan:
I wrote something quick and dirty a few weeks back to encode the form into JSON in order to save the form at the back end using AJAX calls. Let me know if this helps.

`        fjson += '{"questions": [';
        var qs = "";
        $("div.frmb-wrap ul.frmb li.form-field").each(function(){
          var q = ""
          //T Or F (TRF)
          if ($(this).hasClass("checkbox")){q='{"title": "'+ validQCheck($(this).find("input.fld-label").val()) + '", "qtype": "TRF"}, ';}
          //Rating (GRS)
          else if ($(this).hasClass("radio-group")){q='{"title": "'+ validQCheck($(this).find("input.fld-label").val()) + '", "qtype": "GRS"}, ';}
          //TXT
          else if ($(this).hasClass("text")){q='{"title": "'+ validQCheck($(this).find("input.fld-label").val()) + '", "qtype": "TXT"}, ';}
          //MS
          else if ($(this).hasClass("checkbox-group")){
            q='{"title": "'+ validQCheck($(this).find("input.fld-label").val()) + '", "qtype": "MS",';
            q+='"detail": "'
            var res = "";
            $(this).find('input.option-label').each(function(){res += ''+ validCheck($(this).val()) + ',';});
            q += res.substring(0,(res.length -1)) + '" }, ';
          }
          //MC
          else if ($(this).hasClass("select")){
            q='{"title": "'+ validQCheck($(this).find("input.fld-label").val()) + '", "qtype": "MC",';
            q+='"detail": "'
            var res = "";
            $(this).find('input.option-label').each(function(){res += ''+ validCheck($(this).val()) + ',';});
[...]

PR #201, opened 2016-07-18
Implemented saving method toJSON

Yc-Chen:
related to issue #184


kevinchappell:
Looking good. going to test now.

kevinchappell:
I am having difficulty getting this to work. Can you include the `getJSON` function you mentioned in #184 and possibly provide an example on using `toJSON`?

Yc-Chen:
I have created a demo under the 'json' branch here:
[https://github.com/Yc-Chen/formBuilder/tree/json/demo-json](https://github.com/Yc-Chen/formBuilder/tree/json/demo-json)
I do not know how to use gulp so I directly edited the built js file. But I commented 'REVIEW' where I made changes. 
I also implemented the `getJSON` method but it is not working.

A screenshot to show the output json file looks like this:
screenshot - 20-07-16 - 17 36 56

kevinchappell:
Thats a good start. The field objects are missing the name data which may be why its not working but I'll look closer to be sure. Also we'll need a trimming function to keep remove empty properties. There is already something like this in formBuilder called `_helpers.trimAttrs()` but it should really be renamed `trimObj`.

Good work

Yc-Chen:
I figured out what went wrong. Now my master branch should contain the correct `to-json.js` to generate json string. How does this work? Should I send another pull request?

kevinchappell:
Excellent, I'll check it out. You should only need to rebase against master to catch your feature branch up with changes since opening the pr.

kevinchappell:
Looks like you got the branch caught but I'm still getting an error: `getJSON is undefined`.

Yc-Chen:
I tried to break the entry of `getJSON` but it still doesn't work. So I tried to get where `getJSON` is referred. 

`grep -rl "getJSON" .` shows:
`./demo/assets/js/form-builder.min.js`
`./demo/assets/js/form-builder.min.js.map`

`grep -rl "getXML" .` shows:
`./demo/assets/js/form-builder.min.js.map`
`./dist/form-builder.js`
`./src/js/form-builder.js`

It looks something wrong with gulp. But for myself, I included the files in the src folder to the html and it works fine...

kevinchappell:
@Yc-Chen I really appreciate the effort that went into this PR but was not able to get it fully working. It did function as guidance and inspiration for #237 and for that I- am I'm sure a good number of formBuilder users, are very grateful. 

JSON Support is now part of formBuilder as of #237 .

PR #206, opened 2016-07-21
Disable scroll on ControlBox fields

jonrcable:
When creating really long forms (20-30 fields) the drag event from the ControlBox auto scrolls to the bottom. This makes it impossible to drop fields into the middle of a form.  Simple fix, minimal impact.


kevinchappell:
Good catch.

PR #210, opened 2016-07-22
Fix the position of the ControlBox

jonrcable:
Fixed the position of the cb-wrap class to allow the ControlBox to follow the page and make editing easier.

Just add a bunch of fields and start scrolling :)
https://jsfiddle.net/jonrcable/57a57wsm/1/

### DIFF
--- a/dist/form-builder.css
+++ b/dist/form-builder.css
@@ -139,6 +139,8 @@ Author: Kevin Chappell <kevin.b.chappell@gmail.com>
 
 .cb-wrap {
   float: right;
+  position: fixed;
+  right: 5px;
   width: 26%;
   -webkit-transition: -webkit-transform 250ms;
   transition: -webkit-transform 250ms;


kevinchappell:
This is good but css only has a couple issues. Fixed position with `right: 5px` alone may have undesired positioning and width in some layouts. Also if there is content below the editor the control bar will continue to follow to the end of the page. 

For these reasons I went with a hybrid css and js solution for the ‘[stickyControls](http://formbuilder.readthedocs.io/en/latest/formBuilder/options/stickyControls/)‘ option.

kevinchappell:
I forked and updated the fiddle to better illustrate what I meant in my comment: https://jsfiddle.net/kevinchappell/ajp60dzk/2/

Here is what the `stickyControls` option looks like in the same fiddle
https://jsfiddle.net/kevinchappell/zb6mcy4m/1/

I think my implementation could still use some work though since its setting the fixed position with Javascript instead of applying a class. window scroll position test could also be throttled but i'm guessing very few people use this option so will hold off on refactoring until performance issues are flagged.

kevinchappell:
@jonrcable I prefer your css only solution if we could work out the issues I mentioned above but I'm going to close this for now though until we can figure that out.

PR #235, opened 2016-08-28
Added first revision of Dutch language (nl_NL)

xoniq:


kevinchappell:
Fantastic! I will have to add language selection shortly to take advantage of all these languages.

PR #244, opened 2016-09-12
Add new option 'showCommandButtons'

srachner:
I've added a new option which allows deactivating the clear, preview and save buttons in the controls. I need this for a project which already has separate controls for some actions. I also made the preview dialog action accessible, so that all three commands can be triggered from an external script.


kevinchappell:
This is a great PR, everything is where is should be and no linting errors, excellent work @srachner . 

My only 2 questions are if `showCommandButtons` could be `showActionButtons`, and thats only because its referred to as `actions` and `formActions` in the jQuery `data()` and code respectively.

Also maybe `showXML` should be `showData` since formBuilder now has JSON support and `showData` could refer to XML _or_ JSON.

Seriously, fantastic PR.

dist/form-builder.js
```diff
@@ -1510,6 +1519,7 @@ function formBuilderEventsFn() {
       },
       sortableControls: false,
       stickyControls: false,
+      showActionButtons: true,
```
kevinchappell: perfect

srachner:
Thank you! 

You're right about the naming, I've changed it according to your suggestions .

PR #245, opened 2016-09-12
Bugfix: ensure there are field for clearFields action and update demo

kevinchappell:
Added:
`modalOpened` event

Changed:
check for fields before running clearFields action
keep `formActions` in scope


src/js/form-builder.js
```diff
@@ -432,7 +408,32 @@
...
+        saveAll = utils.markup('button', opts.messages.save, {
+          className: `btn btn-primary ${opts.prefix}save`,
+          id: frmbID + '-save',
+          type: 'button'
+        }),
+        formActions = utils.markup('div', [clearAll, viewData, saveAll], {
+          className: 'form-actions btn-group'
+        });
+
+      cbWrap.append(formActions);
```
kevinchappell: This ensures `formActions` is within scope and is only appended if it exists

src/js/helpers.js
```diff
@@ -791,6 +791,10 @@ function formBuilderHelpersFn(opts, formBuilder) {
     var $fields = $(fields);
     var markEmptyArray = [];
 
+    if (!fields.length) {
```
kevinchappell: dump out of the function if there are no fields to clear

demo/assets/js/demo.js
```diff
@@ -9,7 +9,9 @@ jQuery(document).ready(function($) {
     editing = !editing;
   };
 
-  $(buildWrap).formBuilder();
+  var formBuilder = $(buildWrap).formBuilder().data('formBuilder');
+
+  console.log(formBuilder);
```
kevinchappell: give developers quick view of available actions, data etc

PR #246, opened 2016-09-12
Enable HTML in field labels

kevinchappell:
Allow HTML in labels and some elements

minor update to docs


kevinchappell:
Todo: 
Block elements should be converted to `contenteditable` divs
Implement style formatter
possible markdown support

PR #252, opened 2016-09-16
Copy Button Added

jcacchio:
Included new field action button, as well as function for cloning the current field.

For instance, if a form designer wishes to create a survey with multiple "agree/disagree" type questions, they can copy a field and will not have to re-create the field options each time.


src/js/form-builder.js
```diff
@@ -1024,6 +1031,44 @@
...
+                var newVal = this.getAttribute('value');
+                newVal = newVal.substring(0, (newVal.lastIndexOf('-') + 1));
+                newVal = newVal + ts.toString();
+                this.setAttribute('value', newVal);
+            }
+        });
+
+        $clone.attr("id", lastID);
+        $clone.attr("name", cloneName);
+        $clone.addClass("cloned");
```
kevinchappell: Cloned fields with options like select and checkbox groups will need to have their options made sortable again. Adding `$('.sortable-options', $clone).sortable();` below this line seems to fix that.

kevinchappell:
Excellent and works well. Only thing that needs correcting is the comment I left.

src/js/form-builder.js
```diff
@@ -927,10 +929,15 @@
...
+        }),
+        copyBtn = utils.markup('a', opts.messages.copyButton, {
+          id: lastID + '-copy',
+          className: 'copy-button btn icon-copy',
+          title: opts.messages.copyButtonTooltip
         });
 
       var liContents = utils.markup(
-        'div', [toggleBtn, delBtn], { className: 'field-actions' }
+        'div', [toggleBtn, copyBtn, delBtn], { className: 'field-actions' }
```
kevinchappell: Should we change order of action buttons? I think going from most used to least used from right to left makes sense but that would mean moving the delete button also. What are your thoughts? I ask mainly because I keep cloning fields when i mean to edit them.

src/js/form-builder.js
jcacchio: Thanks Kevin.  I did toy with that this morning and pushed that change.

src/js/form-builder.js
jcacchio: I personally don't have a preference.  But anyone currently using formBuilder could be thrown off by flip-flopping edit and delete.

jcacchio: On second thought, I do like having delete in the upper right, since the behavior is similar to "close".

kevinchappell: OK so then the question is should it be `[toggleBtn, copyBtn, delBtn]` or `[copyBtn, toggleBtn, delBtn]`?

I'm fine with it either way.

jcacchio: In that case, let's leave it at `[toggleBtn, copyBtn, delBtn]` keeping the edit button on the left -- closest to the field attributes that need to be edited.

PR #254, opened 2016-09-16
Simple solution for overriding text,number and boolean predefined attrs

gimigliano:
```
typeUserAttrs:{
               'text':{
                       'maxlength':{'type':'number','min':3,'max':10,'step':1},
                      }
}
```


kevinchappell:
I like it, nice and simple.

PR #260, opened 2016-09-27
Added typeUserEvents option

gimigliano:
Suppose you want add an attribute "pattern" in text fields and you want to hide (and disable) it when the "email" subtype is selected, you'll need to attache onchane evento when the text field is added.

     

```
typeUserAttrs:{
                       'text':{
                                'pattern':{'label':'Pattern',
                               }
                       } , 
 typeUserEvents:{
                       'text':{
                               'onadd':function(fld){ fld.find('.fld-subtype:first')
                                                                    .change(function(h){
                                                                                     if($(this).val()=='Password') 
                                                                                            { fld.find('.pattern-wrap:eq(0)').hide();
                                                                                              fld.find('.fld-pattern:eq(0)').prop('disabled',true);
                                                                                            }
                                                                                    else {fld.find('.pattern-wrap:eq(0)').show();
                                                                                            fld.find('.fld-pattern:eq(0)').prop('disabled',false);
                                                                                           }     
                                                                         } );  
                                                               }
                                  }   
                        },
```

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -6,6 +6,7 @@
 
     var defaults = {
       typeUserAttrs: {}, //+gimigliano
+      typeUserEvents: {}, //+gimigliano
       controlPosition: 'right',
       controlOrder: [
         'autocomplete',
@@ -1072,7 +1073,10 @@
         _helpers.closeAllEdit($sortableFields);
         _helpers.toggleEdit(lastID);
       }
-
+      
+      //+gimigliano
+      if (opts.typeUserEvents[type] && opts.typeUserEvents[type]['onadd']) opts.typeUserEvents[type]['onadd']($('#'+lastID));
+     
       lastID = _helpers.incrementId(lastID);
     };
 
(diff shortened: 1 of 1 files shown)


gimigliano:
For each type of field you can add "onadd" and "onclone" event

kevinchappell:
Another great feature!

src/js/form-builder.js
```diff
@@ -1161,6 +1165,9 @@
       $clone.attr('name', cloneName);
       $clone.addClass('cloned');
       $('.sortable-options', $clone).sortable();
+    //+gimigliano
+      if (opts.typeUserEvents[type] && opts.typeUserEvents[type]['onclone']) opts.typeUserEvents[type]['onclone']($('#'+lastID));
```
kevinchappell: I did make a couple small changes after merging.

1. `type` was undefined here so I defined it above
2. `onclone` and `onadd` now receive the JS DOM object instead of performing jQuery lookup on ID. There may uses where developer needs only value or something, and passing existing reference is slightly faster.

Everything else is the same. This is very useful. :+1:

gimigliano: 1. Cut & paste ;
2. OK

PR #276, opened 2016-10-11
Feature: addField and removeField actions

### LINKED ISSUE #269: append field (json string) from database to current formBuilder formData
I'm building forms where the user can also include questions already saved in a database as json strings (from jQuery formBuilder). I'm trying to append these db based json strings into the existing formData. 

I've experimented with concatenating the existing formBuilder.formData json string with the additional dbField json string, then deleting the existing instance of formBuilder and starting a new one (using the new concatenated formData string). this kind of works in an ugly way. 

I can see methods in form-builder.js that append new fields but they are not exposed for external use. is there a way to access these, or another way to push new fields into the existing formData?

and BTW formBuilder is a very impressive project - thanks for your work on it.

kevinchappell:
Fields can now be added and remove programmatically with `addField` and `removeField`.

Resolves #269


mcbmcb0:
thanks for this feature. I'm calling it  like this:
`formBuilder.actions.addField(data, position); //where data is a json object, not string`

one possible glitch: after the field is added it displays ok in the builder but is not in formBuilder.formData untl the form is saved: i think until  _helpers.prepData has done its magic.
is this a glitch or am i missing something?

thanks

kevinchappell:
You're absolutely right. Save should happen after the field is added. Until I add it you should be able to call `actions.save()` right after `actions.addField`

mcbmcb0:
thanks - didn't think of that. that works

PR #291, opened 2016-10-17
Hotfix: deleteId undefined

### LINKED ISSUE #290: fieldRemoveWarn: true - error and won't remove field on 'yes'
hello. 
with v1.21.3, when i set fieldRemoveWarn: true, then click 'yes' i get an error. firebug says:
`TypeError: can't convert undefined to object` at this line: `var o=[].slice.call().map(function(e){return e....`
which appears to be line 1285 of the non-minified version:
      `var availableIds = [].slice.call().map(function (field) {`

this happens on my development site and on the `http://formbuilder.readthedocs.io` at 'Docs » formBuilder Options » fieldRemoveWarn' here:
[(http://formbuilder.readthedocs.io/en/latest/formBuilder/options/fieldRemoveWarn/)]

i think its a bug, but maybe not? not sure how to proceed on this one.
thanks
Mike

kevinchappell:
Resolves #290

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1332,7 +1332,7 @@
       if (opts.fieldRemoveWarn) {
         let warnH3 = utils.markup('h3', opts.messages.warning),
           warnMessage = utils.markup('p', opts.messages.fieldRemoveWarning);
-        _helpers.confirm([warnH3, warnMessage], deleteID => _helpers.removeField(deleteID), coords);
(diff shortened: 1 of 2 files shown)


mcbmcb0:
thanks for the fix - fieldRemoveWarn:true now working but controlOrder now appears not to be:
on docs site and my local:
[(http://formbuilder.readthedocs.io/en/latest/formBuilder/options/controlOrder/)]
?

kevinchappell:
@mcbmcb0 good catch. fixing now.

PR #340, opened 2016-12-12
Repair bower.json

SHBelsky:
As it stands, the bower.json file points to the incorrect bower repo. I've repaired the reference to the bower repo so that installations with bower will not cause an error.

[Bower URL](https://libraries.io/bower/jquery-ui-sortable)


kevinchappell:
Good fix.

PR #344, opened 2016-12-18
Update options.md

erop:
Looks like we need to add dataType to option object to be able to consume formdata in JSON format


kevinchappell:
:+1:

PR #352, opened 2017-01-07
make check on select which is required work

didilem:
with out the empty value a select which has no user  selection the check passes, same applies in form-builder.js


kevinchappell:
Good catch.

PR #354, opened 2017-01-12
fbUtils.parseXML: Check that fields has children

ivuorinen:
Fixes bug in some IE versions where if children were not defined caused
form not being able to render.


kevinchappell:
Nice, good fix.

PR #361, opened 2017-01-24
inputSets name attribute problem

HayriCan:
On line 1965  it was't taking the name attribute from inputSets . It was using nameAttr(field) in all cases.

### DIFF
--- a/dist/form-builder.js
+++ b/dist/form-builder.js
@@ -1961,8 +1961,8 @@ function formBuilderEventsFn() {
       } else {
         field = Object.assign({}, $field);
       }
-
-      field.name = isNew ? nameAttr(field) : field.name || nameAttr(field);
+      
+      field.name = isNew ? field.name || nameAttr(field) : nameAttr(field);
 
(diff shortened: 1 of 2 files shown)


ivuorinen:
You should make your changes to `src` files. The `dist` folder contents are build from `src` and your patch gets overwritten.

dist/form-builder.js
```diff
@@ -1,2905 +1,2905 @@
...
+};
+
+/**
+ * convert a hyphenated string to camelCase
+ * @param  {String} str
+ * @return {String}
+ */
+fbUtils.camelCase = function (str) {
+  return str.replace(/-([a-z])/g, function (m, w) {
+    return w.toUp
```
kevinchappell: maybe we should just change this line to:
```javascript
if(!field.name){
  field.name = nameAttr(field);
}
```

ALso this change would need to be added to the src file [here](https://github.com/HayriCan/formBuilder/blob/master/src/js/form-builder.js#L535)

PR #447, opened 2017-05-10
Update select.js for form-render to work with http post data

MrGate:
This fixes the issue with checkbox groups that have more then one option checked, it gives each input a unique name so all the data can be submitted :) -first git push


MrGate:
Example Before Fix
  'checkbox-group-1494393170194' => string 'Hello-3'

Example After Fix
  'checkbox-group-1494393170194-0' => string 'option-1'
  'checkbox-group-1494393170194-2' => string 'Hello-3'

MrGate:
also tested on my side, working great. i know the default method to name them probably needs to be taken out, but i'm not deep enough in to be able to find it yet!

kevinchappell:
@MrGate Can you confirm if this is still needed? I'm not able to reproduce in 2.2.2.

PR #501, opened 2017-05-26
Hotfix: preload values to exisitng field types, fix fieldOrder

### LINKED ISSUE #275: hide field options & set default values
There is no way to remove field options (e.g. button name or button type).
Set default values to these options would be also good, e.g. the button default style could be primary.

### LINKED ISSUE #377: Charge select options with json
How to load the options of a select based on catalogs, example: show the option to load a select with countries or states, the user selects one of these and the select loads the options based on the corresponding json.

kevinchappell:
Can now preload values to existing field types set via the `fields` option. Example:

```javascript
  var fields = [
    { 
      type: "text", 
      value: "wubba lubba dub dub" 
    },
    {
      type: "autocomplete",
      label: "Favorite Programming Language",
      values: [
        { label: "SQL" },
        { label: "C#" },
        { label: "JavaScript" },
        { label: "Java" },
        { label: "Python" },
        { label: "C++" },
        { label: "PHP" },
        { label: "Swift" },
        { label: "Ruby" }
      ]
    }
  ];
  $(container).formBuilder({ fields: fields });
```

resolves #377, resolves #275


src/js/form-builder.js
```diff
@@ -98,7 +96,7 @@ const FormBuilder = function(opts, element) {
     }
     let icon = custom.icon || controlClass.icon(type);
     let label = custom.label || controlClass.label(type);
-    let iconClassName = !icon ? custom.iconClassName || `icon-${type}` : '';
+    let iconClassName = !icon ? custom.iconClassName || `icon-${type.replace(/-[\d]{4}$/, '')}` : '';
```
brentkelly: @kevinchappell is this `icon-${type.replace(/-[\d]{4}$/, '')}` there to handle the custom fields?

If so a better way to write this section may be:

```
let iconClassName = '';
if (!icon) {
   icon = custom ? custom.iconClassName || custom.type : `icon-${type}`;
}

Comment (kevinchappell) [inline src/js/form-builder.js]: Agree on the custom type but since since the custom type key is randomly generated each time, even if there was an icon of say 'autocomplete-2847' those last 4 digits would change on each initialization. The regex removes hyphen and 4 digits which should only be generated by custom fields. If the custom field is actually an extended base type like in the autocomplete example, its possible the autocomplete icon still wants to be used. To make that happen we strip the '-2847' leaving 'icon-autocomplete'. If they don't want the autocomplete icon they can always explicitly define an icon or iconClassName.

Comment (brentkelly) [inline src/js/form-builder.js]: Yep I guess either way works. My implementation above removes ignores the '-1234' by just using the original type defined in `opts.fields`. So
```
var fields [
  {
    type => 'text',
      ...
  }
];
```

... would then output `icon-text` for that custom field.

Either way should have the same outcome so up to you which you'd prefer.

Comment (kevinchappell) [inline src/js/form-builder.js]: Ah ok. We can always change it if it becomes an issue.

File: src/js/form-builder.js [hunk 2]
```diff
@@ -108,27 +106,37 @@ const FormBuilder = function(opts, element) {
...
-      $(inputSet).appendTo($cbUL);
+      inputSet.dataset.type = set.name;
+      controlList.push(set.name);
+      allControls[set.name] = inputSet;
     });
   }
 
+  // remove disableFields
+  if (opts.disableFields) {
+    controls = controls.filter(type => opts.disableFields.indexOf(type) == -1);
```
aromka: this should go either before `for` loop above, or filter out controlList /  allControls

kevinchappell: moved to orderFields in helpers

PR #509, opened 2017-05-29
upgraded fineuploader plugin to use cdnjs by default so it no longer …

brentkelly:
…requires installation, make template configurable

Thanks @rnicholus for pointing out the CDN.

@kevinchappell - one thing you may want to consider now that it doesn't require installation & 'just works', is if it is worth including this in the core set of controls - either as the default `file uploader` control, or possibly in the `subtype` dropdown for `file uploader`. 

Alternatively can just leave it as a plugin - which would make sense if a lot of the users wouldn't use this functionality.

Guess it depends how lightweight you are trying to keep the project out of the box.

### DIFF
--- a/src/js/control/file.fineuploader.js
+++ b/src/js/control/file.fineuploader.js
@@ -1,3 +1,5 @@
+import controlText from './text';
+
 /**
  * Fineuploader class - render the fineuploader tool (https://fineuploader.com) in place of the traditional file upload widget
  * For assistance with further configuring Fine Uploader in your application, please refer to:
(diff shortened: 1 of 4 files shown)


kevinchappell:
@brentkelly I definitely think its worth it and should exist as a subtype. If a dev wants it to be the default they can either use combination of `disableFields` and `fields` option or the `replaceType` option I am proposing in #510

brentkelly:
OK cool - I'll rework this PR & let you know when it's good to go.

PR #516, opened 2017-06-01
Fineuploader error handling & reporting

brentkelly:
Fineuploader silently fails by default when your server endpoint returns an error (wrong file type, too large, unwriteable upload dir etc). This PR adds support for displaying those errors in a user friendly way within the control. 

To support this I also upgraded controls to be able to define css rules with the `this.css` array (so it now supports both externally linked CSS, as well as inline css rules).

Tested in chrome, edge, and firefox.

There were also a couple of other more obscure bugs in fineuploader that are fixed in this PR.

My formbuilder demo won't compile in IE so couldn't test there, but it's pretty standard jQuery so can't see why it wouldn't work (then again ... we are talking about IE!)

### DIFF
--- a/src/js/control/file.fineuploader.js
+++ b/src/js/control/file.fineuploader.js
@@ -50,105 +50,105 @@ export default class controlFineUploader extends controlText {
    */
   configure() {
     this.js = this.classConfig.js || '//cdnjs.cloudflare.com/ajax/libs/file-uploader/5.14.2/jquery.fine-uploader/jquery.fine-uploader.min.js';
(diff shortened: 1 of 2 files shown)


brentkelly:
@kevinchappell - I added another commit to this PR. As I've started to use this in a project, I noticed it was missing some important features. This latest one is it wasn't actually outputting the names of the files uploaded to a hidden input or anything like that. So to get them, you had to interact with the fineuploader object over javascript etc. Doesn't really fit the pattern of other rich components like the RTE controls which work transparently with a form post using a hidden input. So I've upgraded this to do the same in https://github.com/kevinchappell/formBuilder/pull/516/commits/6660c20640f3317470b7d1fffc548c9c2c2a2f88

PR #521, opened 2017-06-05
fixed autocomplete control behaviour

devstructor:
autocomplete control was not rewriting valules to the hidden field and therefore it was not submitted. Moreover its behaviour was not intuitive. I fixed all of the issues I noticed


kevinchappell:
Really liked these changes. Much more user friendly :+1:

PR #550, opened 2017-07-07
Add support for disabling form action buttons (copy, edit, remove)

mikevoets:
In my case there is a need for static fields that users are not allowed to remove or edit. Therefore I added support for this by allowing developers to specify which field action buttons they want to disable, in the field values, through a disableFieldButtons array containing one or more of ['copy', 'edit', 'remove'].

Whenever disabling the edit field button, the dblclick event on the field will also be ignored.

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1007,24 +1007,38 @@ const FormBuilder = function(opts, element) {
   let appendNewField = function(values, isNew = true) {
     let type = values.type || 'text';
     let label = values.label || i18n[type] || i18n.label;
-    let delBtn = m('a', i18n.remove, {
+    let fieldButtons = [
+      m('a', i18n.remove, {
+        type: 'remove',
         id: 'del_' + data.lastID,
         className: 'del-button btn delete-confirm',
         title: i18n.removeMessage
+      }),
+      m('a', null, {
+        type: 'edit',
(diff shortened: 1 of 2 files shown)


src/js/form-builder.js
```diff
@@ -1007,24 +1007,38 @@ const FormBuilder = function(opts, element) {
...
+      }),
+      m('a', null, {
+        type: 'copy',
+        id: data.lastID + '-copy',
+        className: 'copy-button btn icon-copy',
+        title: i18n.copyButtonTooltip
+      })
+    ];
+
+    let disabledFieldButtons = values.disabledFieldButtons;
```
kevinchappell: since this is coming from `values`, will generated fields have a `disabledFieldButtons` attribute?

src/js/helpers.js
```diff
@@ -782,6 +782,7 @@ export default class Helpers {
   toggleEdit(fieldId, animate = true) {
     const field = document.getElementById(fieldId);
     const toggleBtn = $('.toggle-form', field);
+    if (!toggleBtn.length) return;
```
kevinchappell: 👍

src/js/form-builder.js
mikevoets: Yes.

Issue #612, opened 2017-09-14
Checkbox Group Other option doesn't get a name


craigpotter:
### Description:
If you render a form with a checkbox group and select the option to have a "other" option with a textbox, the "other" field doesn't get name so it's ever submitted to the server. 

### Environment Details:

 - formBuilder Version: 2.9.6

### Expected Behavior
Have the same name as a checkbox.

### Actual Behavior
No name which means the field isn't passed when the form is submitted

### Steps to Reproduce
Create a checkbox group. Select "Other" option. Check out form rendered.

### Screenshot - *(optional)*
![image](https://user-images.githubusercontent.com/1442635/30430300-b48ef9e8-9952-11e7-93bd-4a295e36fcc1.png)

cklone:
@craigpotter I believe that PR #629 should also fix this issue. While it doesn't give the text input (the one you have highlighted) a name attribute, it does update the value of the other input `id: checkbox-150...-other` when the user types. This was the previous behavior before a bug was introduced. The form will now submit the value with the name `checkbox-150...` which is the intended behavior.

kevinchappell:
This seems to be functioning as intended now but what catches my eye is that the checkbox group in your screenshot has the same name for each input instead of `checkbox-15033298729873[]` note the `[]` at the end. Is there an issue with receiving values from checkbox-groups?

craigpotter:
@kevinchappell  It seemed to work when I was testing it. 

I just upgraded to 2.9.8 and the [] are visible. I'm not sure if this was a bug or just a chrome issue. 

Seems fine now.

kevinchappell:
Excellent!

PR #629, opened 2017-09-29
Fixes #628 - Cannot set property 'value' of null (typing value into "other field" after render)

### LINKED ISSUE #628: Uncaught TypeError: Cannot set property 'value' of null (typing value into "other field" after render)
### Description:
Getting a JS console error after entering a value into the "other" text field: `Uncaught TypeError: Cannot set property 'value' of null`.

### Environment Details:

 - formBuilder Version: v2.9.6
 - Browser: Chrome Version 61.0.3163.91
 - OS: Mac OSX Version 10.12.6 (16G29) (Sierra)

### Expected Behavior
That the value get set correctly without a JS error. Potentially related to #612.

### Actual Behavior
JS Console error and value doesn't get set correctly.

### Steps to Reproduce
- Create a checkbox-group and enable "other" option
- Render form
- Check "other" and type in the text field
- Should see JS console error

cklone:


cklone:
Should fix #628 as well as #612.

PR #749, opened 2018-05-12
Bring formbuilder-formrender + form data together

Jojoshua:
@kevinchappell  @wobh

This is my attempt at bringing the formbuilder definition together with formrender plus the data that goes with the fields all together. An additional object called userData is added to the controls which can be saved in a database and pulled back for formrender to populate again. indexRender.html is a new demo page for testing this.

userData works for autocomplete, select, checkbox-group, radio-group, text, email, color, tel, number, hidden,date, textarea, textarea-tinymce.

For fields that have "other" option, a value that is not in the pre-defined values is assumed to be the "other" value.

I intend to use this by calling JSON.stringify(frInstance.userData) before posting to the database.

Other changes:
- Adds a feature to respect the disabled attribute. Also made tinymce respect disabled attribute if set on the textarea.
- Adds a feature to autocomplete which can optionally enforce only selecting the valid pre-defined options
- Minor change to the getStarted text
- Added h4 to paragraph

### DIFF
--- a/demo/assets/js/demoRender.js
+++ b/demo/assets/js/demoRender.js
@@ -57,6 +57,30 @@ jQuery(function($) {
       resetFR();  
     });
 
+      //password
+    document.getElementById('set-password').addEventListener('click', function() {
+      setFormData = '[{"type":"text","label":"Text Field","name":"text-1526099104236","subtype":"password","userData":["Text"]}]';  
+      resetFR();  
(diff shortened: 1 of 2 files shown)


Jojoshua:
Separated the core changes in https://github.com/kevinchappell/formBuilder/pull/750

PR #750, opened 2018-05-13
Bring formbuilder-formrender + form data together

Jojoshua:
@kevinchappell  @wobh --I redid the commits to make it cleaner. Still learning git..

This is my attempt at bringing the formbuilder definition together with formrender plus the data that goes with the fields all together. An additional object called userData is added to the controls which can be saved in a database and pulled back for formrender to populate again. indexRender.html is a new demo page for testing this.

userData works for autocomplete, select, checkbox-group, radio-group, text, email, color, tel, number, hidden,date, textarea, textarea-tinymce.

For fields that have "other" option, a value that is not in the pre-defined values is assumed to be the "other" value.

I intend to use this by calling JSON.stringify(frInstance.userData) before posting to the database.

Other changes:

Adds a feature to respect the disabled attribute. Also made tinymce respect disabled attribute if set on the textarea.
Adds a feature to autocomplete which can optionally enforce only selecting the valid pre-defined options
Minor change to the getStarted text
Added h4 to paragraph

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -7,7 +7,7 @@ export const defaultOptions = {
     'button',
     'checkbox',
     'checkbox-group',
-    'date',    
+    'date',
     'file',
     'header',
     'hidden',
@@ -47,8 +47,8 @@ export const defaultOptions = {
   inputSets: [],
   replaceFields: [],
   roles: {
-    1: 'Administrator'  
-  },  
+    1: 'Administrator'
+  },
   notify: {
     error: message => console.error(message),
     success: message => console.log(message),
@@ -121,7 +121,7 @@ export const defaultI18n = {
       inlineDesc: 'Display {type} inline',
       label: 'Label',
       labelEmpty: 'Field Label cannot be empty',
-      limitRole: 'Limit access to one or more of the following roles:',      
+      limitRole: 'Limit access to one or more of the following roles:',   
       mandatory: 'Mandatory',
       maxlength: 'Max Length',
       minOptionMessage: 'This field requires a minimum of 2 options',
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -7,12 +7,12 @@ export const defaultOptions = {
     'button',
     'checkbox',
     'checkbox-group',
-    'date',
+    'date',    
     'file',
     'header',
     'hidden',
+    'number',    
     'paragraph',
-    'number',
     'radio-group',
     'select',
     'text',
@@ -22,7 +22,7 @@ export const defaultOptions = {
   // Array of fields to disable
   disableFields: [],
   disabledAttrs: [],
-  disabledActionButtons: [],
+  disabledActionButtons: [], // save,data,clear
   disabledFieldButtons: {},
   editOnAdd: false,
   // Uneditable fields or other content you would like to appear
@@ -47,8 +47,8 @@ export const defaultOptions = {
   inputSets: [],
   replaceFields: [],
   roles: {
-    1: 'Administrator',
-  },
+    1: 'Administrator'  
+  },  
   notify: {
     error: message => console.error(message),
     success: message => console.log(message),
@@ -111,9 +111,9 @@ export const defaultI18n = {
       enableOtherMsg: 'Let users to enter an unlisted option',
       fieldNonEditable: 'This field cannot be edited.',
       fieldRemoveWarning: 'Are you sure you want to remove this field?',
-      fileUpload: 'File Upload',
+      fileUpload: 'File Upload',         
       formUpdated: 'Form Updated',
-      getStarted: 'Drag a field from the right to this area',
+      getStarted: 'Drag or click a field from the right to this area',
       header: 'Header',
       hide: 'Edit',
       hidden: 'Hidden Input',
@@ -121,7 +121,7 @@ export const defaultI18n = {
       inlineDesc: 'Display {type} inline',
       label: 'Label',
       labelEmpty: 'Field Label cannot be empty',
-      limitRole: 'Limit access to one or more of the following roles:',
+      limitRole: 'Limit access to one or more of the following roles:',      
       mandatory: 'Mandatory',
       maxlength: 'Max Length',
       minOptionMessage: 'This field requires a minimum of 2 options',
@@ -157,6 +157,7 @@ export const defaultI18n = {
       removeOption: 'Remove Option',
       remove: '&#215;',
       required: 'Required',
+      requireValidOption: 'Only accept a pre-defined Option',     
       richText: 'Rich Text Editor',
       roles: 'Access',
       rows: 'Rows',
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -121,7 +121,7 @@ export const defaultI18n = {
       inlineDesc: 'Display {type} inline',
       label: 'Label',
       labelEmpty: 'Field Label cannot be empty',
-      limitRole: 'Limit access to one or more of the following roles:',   
+      limitRole: 'Limit access to one or more of the following roles:',  
       mandatory: 'Mandatory',
       maxlength: 'Max Length',
       minOptionMessage: 'This field requires a minimum of 2 options',
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -111,7 +111,7 @@ export const defaultI18n = {
       enableOtherMsg: 'Let users to enter an unlisted option',
       fieldNonEditable: 'This field cannot be edited.',
       fieldRemoveWarning: 'Are you sure you want to remove this field?',
-      fileUpload: 'File Upload',         
+      fileUpload: 'File Upload',
       formUpdated: 'Form Updated',
       getStarted: 'Drag or click a field from the right to this area',
       header: 'Header',
--- a/src/js/control/autocomplete.js
+++ b/src/js/control/autocomplete.js
@@ -12,7 +12,7 @@ export default class controlAutocomplete extends control {
    */
   build() {
     let {values, type, ...data} = this.config;
-    const keyboardNav = (e) => {     
+    const keyboardNav = (e) => {
       const list = e.target.nextSibling.nextSibling;
       const hiddenField = e.target.nextSibling;
       let activeOption = this.getActiveOption(list);
--- a/src/js/control/autocomplete.js
+++ b/src/js/control/autocomplete.js
@@ -229,7 +229,7 @@ export default class controlAutocomplete extends control {
    * @return {Object} - is the option in the pre defined list
    */
   isOptionValid(list,value){
-    const options = list.querySelectorAll('li'); 
+    const options = list.querySelectorAll('li');  
     let validValue = false;
     for (let i = 0; i < options.length; i++) {
       if(options[i].innerHTML === value){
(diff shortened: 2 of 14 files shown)


src/js/config.js
```diff
@@ -111,17 +111,17 @@ export const defaultI18n = {
       enableOtherMsg: 'Let users to enter an unlisted option',
       fieldNonEditable: 'This field cannot be edited.',
       fieldRemoveWarning: 'Are you sure you want to remove this field?',
-      fileUpload: 'File Upload',
+      fileUpload: 'File Upload',
```
kevinchappell: lots of white space being added

src/js/config.js
```diff
@@ -111,17 +111,17 @@ export const defaultI18n = {
       enableOtherMsg: 'Let users to enter an unlisted option',
       fieldNonEditable: 'This field cannot be edited.',
       fieldRemoveWarning: 'Are you sure you want to remove this field?',
-      fileUpload: 'File Upload',
+      fileUpload: 'File Upload',         
       formUpdated: 'Form Updated',
-      getStarted: 'Drag a field from the right to this area',
+      getStarted: 'Drag or click a field from the right to this area',
```
kevinchappell: :+1:

src/js/control/autocomplete.js
```diff
@@ -192,22 +211,78 @@ export default class controlAutocomplete extends control {
...
 
   /**
-   * When the element is rendered into the DOM, execute the following code to initialise it
-   * @param {Object} evt - event
+   * Is the value in the autocomplete field in the pre-defined Options list?
+   * @param {Object} list - list of autocomplete options
+   * @param {Object} value -value trying to be set
+   * @return {Object} - is the option in the pre defined list
    */
+  isOptionValid(list,value){
```
kevinchappell: nice!

src/js/control/autocomplete.js
```diff
@@ -192,22 +211,78 @@ export default class controlAutocomplete extends control {
...
+    const options = list.querySelectorAll('li'); 
+    let validValue = false;
+    for (let i = 0; i < options.length; i++) {
```
kevinchappell: minor but could probably shorten this with 
```
return options.some(options => option.innerHTML === value)
```
like loop with `break`, it will return the moment it finds match
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/js/control/autocomplete.js
```diff
@@ -192,22 +211,78 @@ export default class controlAutocomplete extends control {
...
   onRender(evt) {
+    // Set userData if available
+    if(this.config.userData){  
+      let id = $('#'+this.config.name).attr('id');
+      let preSelectedOption = this.config.userData[0];
+ 
+      const list = document.getElementById(id).nextSibling;
+      let selectedOption= null;
+
+      $('#' +id + '-list' + ' li').each(function(){
```
kevinchappell: these selectors can get expensive. maybe we can scope this

src/js/control/autocomplete.js
```diff
@@ -192,22 +211,78 @@ export default class controlAutocomplete extends control {
...
+         return;
+       }else{
+         // Set it to whatever the value is
+        $('#' + id).prev().val(this.config.userData[0]);
+        return;         
+       }
+     }     
+
+      $('#' + id).prev().val(selectedOption.innerHTML);
+      $('#' + id).val(selectedOption.getAttribute('value'));
```
kevinchappell: should maintain a reference instead of multiple query of `$('#' + id)`

src/js/control/autocomplete.js
```diff
@@ -192,22 +211,78 @@ export default class controlAutocomplete extends control {
...
+      
+      if (list.style.display === 'none') {
+        this.showList(list, selectedOption);
```
kevinchappell: where do `showList` and `hideList` come from?

src/js/control/select.js
```diff
@@ -195,6 +195,67 @@ export default class controlSelect extends control {
...
+
+    /**
+   * onRender callback
+   */
+  onRender() {
+    // Set userData if available
+    if(this.config.userData){ 
+      let selectedOptions = [];
+      for (let i = 0; i < this.config.userData.length; i++) {
+        selectedOptions.push(this.config.userData[i]);
```
kevinchappell: could this loop be replace with `selectedOptions = this.config.userData.slice()`?

src/js/control/textarea.tinymce.js
```diff
@@ -22,7 +22,7 @@ export default class controlTinymce extends controlTextarea {
    * configure the tinymce editor requirements
    */
   configure() {
-    this.js = ['//cdn.tinymce.com/4/tinymce.min.js'];
+    // this.js = ['//cdn.tinymce.com/4/tinymce.min.js'];
```
kevinchappell: ?

src/js/control/textarea.tinymce.js
```diff
@@ -60,6 +60,10 @@ export default class controlTinymce extends controlTextarea {
   build() {
     let {value = '', ...attrs} = this.config;
     this.field = this.markup('textarea', this.parsedHtml(value), attrs);
+    // Make the editor read only if disabled is set on the textarea
+    if(attrs.disabled){
+      this.editorOptions.readonly = true;
```
kevinchappell: :tada:

src/js/form-render.js
```diff
@@ -240,10 +240,79 @@ class FormRender {
...
+          if(definitionField.disabled)
+            continue;
+          
+          // Pull all data for the definition
+          let userData = [];        
+          let foundData = false;
+          for (let j = 0; j < userDataFields.length; j++){            
+            if(definitionField.name == userDataFields[j].name){
+              foundData = true;
+              // var test = userDataFields[j].value.replace(/'/g, "&#x27;").replace(/"/g, '&#x22;').replace(/\n/g, "\\n").replace(/\r/g, "\\r");
```
kevinchappell: dont forget to remove these

src/js/control/autocomplete.js
Jojoshua: These already existed further up in the file. ~ lines 150-170

src/js/control/autocomplete.js
Jojoshua: I tried, but options does not allow the .some method.

src/js/control/textarea.tinymce.js
Jojoshua: This took a long time to load so I pulled it out because I already was referencing it in my other project. How can I tell FB to not attempt to load this?

src/js/control/autocomplete.js
kevinchappell: ah yeah youre right because its an HTMLCollection and not an array.

src/js/control/autocomplete.js
kevinchappell: ah ok. i didnt see in PR diff and its been a while since i wrote that :smile:

Jojoshua: Ok I updated this. Seems like it is more performant

src/js/control/select.js
Jojoshua: Yes I did slice(0) and it seems to work fine.

src/js/config.js
Jojoshua: Yeah I tried to fix all of these in the last update.

src/js/form-render.js
Jojoshua: Done

src/js/control/autocomplete.js
Jojoshua: Check it out now and see if it is better.

kevinchappell:
Great work. I've merged but would you mind opening a second PR for documentation?

Jojoshua:
@kevinchappell  Thanks a lot. Here is https://github.com/kevinchappell/formBuilder/pull/755 for some documentation on how this would be used.

jaywelj:
image
I kept on getting this message uppon calling userData :(

amiqat:
image

Jojoshua:
Did you run the indexRender.html demo? yarn start it and navigate there.

marcus-at-localhorst:
Is there a way to get a compiled js file with this feature?

BrianHVB:
@marcus-at-localhorst I'm not sure what you mean by compiled. If you are are referring to a minified .js file, then you can generate one yourself very easily. Here are the steps:

1. Create a new directory on your computer and `cd` into it.
2. `git clone https://github.com/kevinchappell/formBuilder.git`
3. Enter the local git directory
4. Checkout the branch with the latest features (currently *release/2.10.0*)
   * `git checkout release/2.10.0`
5. Install dependencies
   * `npm install`
6. Build the minified files
   * `npm run build`
7. The files will be outputted to the *./dist/* directory.

BrianHVB:
@Jojoshua 

Did you test with the `userData` getter with all types of formBuilder controls? As of 477f6fe220830cee87de61318c5d02edc1cbb951 the getter doesn't seem to be working for the grouped controls (`type = "checkbox-group", type = "radio-group"`). The `userData` array for these types of fields contains only a single `""` entry. Based on the `userData` output from the other controls, my assumption is that the array should contain a list of values that correspond to the checked or selected items.

I didn't find any mention of `checked` in the new code added to *form-render* and I think this is the only way to capture the selected state of the controls. (I also didn't look super close at the code). I ended up making my own function that will scan the form-rooted DOM for group controls that correspond to the control name, and if any are checked, add their value to the `userData` list. Here is the code:

```javascript
getFixedUserData() {
	const ud = this.formViewer.userData;

	// only operate on controls that could have multiple values selected
	ud.filter(control => control.type.includes('group'))
		.forEach(control => {
			// the array already has an empty string in it, so we need to clear
			control.userData = [];
			// each data entry has a list of values, but no proprty to indicate if any of those values are checked
			
			control.values.forEach((itm, index) => {
				// inputs for grouped names use the format groupName-0, groupName-1, ...
				const $input = $(`#${control.name}-${index}`);

				// if the input is checked, add it's value to the userData list.
				if ($input.is(':checked')) {
					control.userData.push(itm.value);
				}
			})
		});
	return ud;
}

I've only tested this with my current project, which uses a subset of the available controls, but looking at the source code, I think it should work. I also didn't test on controls where the *id* was set by the user (instead of being auto-generated), but I don't think that will affect the naming scheme.

Kind of a clumsy hack, but I wanted to hold off on anything more until I got some input on the issue.
```

BrianHVB:
Untested fix for https://github.com/kevinchappell/formBuilder/pull/750#issuecomment-426801472

```javascript

  clear() {
    this.instanceContainers.forEach(container => {
      // clear tinyMCE editors
      this.options.formData
        .slice()
        .filter(fieldData => fieldData.subtype === 'tinymce')
        .forEach(fieldData => window.tinymce.get(fieldData.name).setContent(''))
      container.querySelectorAll('input, select, textarea').forEach(input => {
        if (input.type === 'checkbox' || input.type === 'radio') {
           input.checked = false;
        }
       else {
          input.value = "";
       }
    })
  }
```

Jojoshua:
@BrianHVB Thanks for taking a look at this. As you saw I aimed to support all FormBuilder controls and even TinyMCE. I did not try using the clear() method and did not test how it affects userData. In fact...I was looking through my cloned repo and can't even find clear(). Where is it? Your fix looks reasonable.

BrianHVB:
Thanks. I'll test the fix on an independent sample tomorrow and will make a PR if it holds up.

As for the `clear()`, its at the bottom of [form-render.js](https://github.com/kevinchappell/formBuilder/blob/7c2ec44cfda0840a68ab01f5ba5c4d870c3d897e/src/js/form-render.js#L274-L303)

https://github.com/kevinchappell/formBuilder/blob/7c2ec44cfda0840a68ab01f5ba5c4d870c3d897e/src/js/form-render.js#L274-L303

Jojoshua:
Now I see...looks like Kevin made some updates to userData after I did. clear() did not exist in my repo.

PR #755, opened 2018-05-27
User data docs

Jojoshua:
Docs for userData


jaywelj:
I hope we could get examples on how to use this :( i cant get it in use even after following the documents.

Issue #736, opened 2018-04-23
the name of a multi file upload should be an array 


ArnoutCreemers:
### Description:
When creating a multi-file uploader with the formbuilder plugin, the name of the file input element is not an array. This causes issues when processing the uploaded files.

### Steps to reproduce:

1. Go to https://formbuilder.online/
2. Add a file uploader
3. Go to settings and check "Allow users to upload multiple files"
4. Save the form, copy the HTML

### Expected Behavior
The input file is generated as:
`<input type="file" class="form-control" name="file-1524506075088[]" multiple="true" id="file-1524506075088">`

Note the **[]** behind the name of the element

### Actual Behavior
The input file is generated as:
`<input type="file" class="form-control" name="file-1524506075088" multiple="true" id="file-1524506075088">`

ArnoutCreemers:
For those with the same issue, a dirty fix is to execute following jQuery after the form is rendered:
` $('input[type="file"][multiple="true"]').each(function() {
                $(this).attr("name", $(this).attr("name") + "[]");
            });`

sahilr2050:
@ArnoutCreemers How do I save the uploaded file from the form which is made from FormBuilder? 

> I have made the form using form builder
> I have rendered the form
> User now upload the file and click on submit. I can see the file-xxxx in the post but how can I save the file into the server and in database?

sahilr2050:
I have latest version of Form Builder Still the if input files allow multiple files then name should be an array. Still this is simple name

Spl4x:
> For those with the same issue, a dirty fix is to execute following jQuery after the form is rendered:
> 
> ```
> $('input[type="file"][multiple="true"]').each(function() {
>      $(this).attr("name", $(this).attr("name") + "[]");
>  });
> ```

Great, it helped me rename several fields!

PR #780, opened 2018-07-21
[FIX] Format Name Attr correctly

berniechiu:
### Description

This might resolve https://github.com/kevinchappell/formBuilder/issues/736

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -317,9 +317,7 @@ const FormBuilder = function(opts, element) {
       field = Object.assign({}, $field)
     }
 
-    if (!field.name) {
-      field.name = utils.nameAttr(field)
-    }
+    field.name = utils.nameAttr(field)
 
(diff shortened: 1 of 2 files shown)


src/js/form-builder.js
```diff
@@ -317,9 +317,7 @@ const FormBuilder = function(opts, element) {
       field = Object.assign({}, $field)
     }
 
-    if (!field.name) {
```
kevinchappell: Without this check, what happens when saved data is loaded? will it replace the name?

berniechiu: Actually, I found `utils.nameAttr()` output to be the same as `field.name` in most cases, maybe I'll look into more. Because I thought the `FormBuilder` generated that value automatically.

kevinchappell: It does generate it if one isn't set. My concern was that if an existing form was loaded, the saved value could be overwritten. If a user sets a custom name, that custom value would be overwritten.

berniechiu: Cool, i've check the field name inside name utility function

PR #828, opened 2018-10-04
Fix bug with clear() as reported in #750

BrianHVB:
Add control type check to clear() so that method doesn't reset the value for checkboxes and radio buttons, which caused the userData getter to work incorrectly.

See https://github.com/kevinchappell/formBuilder/pull/750#issuecomment-426801472 for more.


kevinchappell:
Good catch. merging.

PR #840, opened 2018-10-30
Store starRating selection in a hidden element so it can be submitted with the form as normal post data

pfranza:
Modify the starRating plugin to back the rateYo object with a hidden input field, so the field can be sent as a normal form POST.


kevinchappell:
I was just about to test this. Just curious why you closed PR?

PR #841, opened 2018-10-30
Back the star rating plugin with a hidden input field so it can behave as a normal form POST object

pfranza:


src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
...
     }
 
     /**
      * onRender callback
      */
     onRender() {
       let value = this.config.value || 3.6;
-      $('#'+this.config.name).rateYo({rating: value});
+      var targetId = this.config.name;
+      $('#' + this.config.name + "_render").rateYo({
```
kevinchappell:
```suggestion
      $(this.field).rateYo({
```
I realize this had extra lookup before but while its being updated we can use `this.field` reference instead to avoid additional jQuery lookup. It will also fix the mix of single and double quotes.

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
...
+          rating: value,
+          onSet: function onSet(rating, rateYoInstance) {
+            $('#' + targetId).val(rating);
```
kevinchappell:
```suggestion
            $(this.input).val(rating);
```

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
...
+      return this.markup('div', [this.input, this.field]);
     }
 
     /**
      * onRender callback
      */
     onRender() {
       let value = this.config.value || 3.6;
-      $('#'+this.config.name).rateYo({rating: value});
+      var targetId = this.config.name;
```
kevinchappell: not needed if using `this.input` assigned in `build()` method, suggestion below.

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
      * @return {Object} DOM Element to be injected into the form.
      */
     build() {
-      return this.markup('span', null, {id: this.config.name});
+      this.inputId = this.config.name;
+      let defaultValue = this.config.value || 3.6;
```
kevinchappell: `defaultValue` is never reassigned could use const

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
...
+      this.input = this.markup('input', null, {
+        name: this.inputId,
+        id: this.inputId,
+        value: defaultValue,
```
kevinchappell: `defaultValue` isn't used anywhere else so I think this could just be
```suggestion
        value: this.config.value || 3.6,
```

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
      * @return {Object} DOM Element to be injected into the form.
      */
     build() {
-      return this.markup('span', null, {id: this.config.name});
+      this.inputId = this.config.name;
```
kevinchappell: `this.inputId` isn't used elsewhere in the class, no need to attach.

src/js/control_plugins/starRating.js
```diff
@@ -36,15 +36,32 @@ window.fbControls.push(function(controlClass) {
...
     /**
      * onRender callback
      */
     onRender() {
       let value = this.config.value || 3.6;
-      $('#'+this.config.name).rateYo({rating: value});
+      var targetId = this.config.name;
+      $('#' + this.config.name + "_render").rateYo({
+          rating: value,
+          onSet: function onSet(rating, rateYoInstance) {
```
kevinchappell: `rateYoInstance` is unused

kevinchappell:
I'd like to merge this. @pfranza can you rebase to resolve the conflicts? Note that `dist/` was removed from repo.

PR #859, opened 2018-11-07
bootstrap grid/column support

Jojoshua:
There are a lot of requests for column support. Here is my implementation. I haven't tested with all form elements but the ones I needed worked fine.

This introduces the concept of columns using bootstrap grid system via class attributes set on the formBuilder elements. To set this up ....

1. add a row identifier onto the class i.e row-1
2. add the bootstrap grid columns i.e col-xs-6

All elements with the same row- identifier will be grouped into the same row
A second row can be added by repeating the same steps but setting a different row- identifier i.e row-2

### DIFF
--- a/src/js/form-render.js
+++ b/src/js/form-render.js
@@ -86,18 +86,14 @@ class FormRender {
         fields.forEach(field => {
           // Determine if rows are being used. If so, create the row and append to its row-{group}
           // If the fields have row-, create & append to the appropriate row
-          const classes = field.className.split('row-');
-          if(classes[1]){           
-            const rowGroup = classes[1].split(' ')[0];
+          const [ rowGroup ] = field.className.match(/row-([^\s]+)/)
+          if(rowGroup){           
             const rowID = this.id ? `${this.id}-row-${rowGroup}` : `row-${rowGroup}`;
 
             // Check if this rowID is created yet or not.            
             let rowGroupNode = document.getElementById(rowID);
-            if(!rowGroupNode){              
-              rowGroupNode = document.createElement('div');
-              rowGroupNode.id = rowID;
-              rowGroupNode.classList.add('row');   
-              rowGroupNode.classList.add('form-inline');
+            if(!rowGroupNode){ 
+              rowGroupNode = utils.markup('div', null, { id: rowID, className: 'row form-inline' });
               renderedFormWrap.appendChild(rowGroupNode)
             }
             rowGroupNode.appendChild(field);              
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -31,7 +31,7 @@ export default class layout {
     
         // Lift any col- and row- type class to the form-group wrapper. The row- class denotes the row group it should go to
         if(data.className){
-          const classes = data.className.split(' ').filter(className => (/^col-(xs|sm|md|lg)-\d+/.test(className) || className.startsWith('row-')));
+          const classes = data.className.split(' ').filter(className => (/^col-(xs|sm|md|lg)-([^\s]+)/.test(className) || className.startsWith('row-')));
           if(classes && classes.length > 0){
             className += ` ${classes.join(' ')}`;
 


src/js/form-render.js
```diff
@@ -82,10 +82,34 @@ class FormRender {
...
-          field.dispatchEvent(events.fieldRendered)
+          // Determine if rows are being used. If so, create the row and append to its row-{group}
+          // If the fields have row-, create & append to the appropriate row
+          const classes = field.className.split('row-');
+          if(classes[1]){           
+            const rowGroup = classes[1].split(' ')[0];
+            const rowID = this.id ? `${this.id}-row-${rowGroup}` : `row-${rowGroup}`;
+
+            // Check if this rowID is created yet or not.            
+            let rowGroupNode = document.getElementById(rowID);
```
kevinchappell: Can use `utils.markup` and avoid if block on [L96](https://github.com/kevinchappell/formBuilder/pull/859/files#diff-47035f04c42bc9d79ae2e3e19a6eb04fR96)
```suggestion
            const rowGroupNode = document.getElementById(rowID) || utils.markup('div', null, { id: rowID, className: 'row form-inline' });
```

src/js/layout.js
```diff
@@ -27,7 +27,22 @@ export default class layout {
...
-        const className = data.id ? `fb-${data.type} form-group field-${data.id}` : ''
+        let className = data.id ? `fb-${data.type} form-group field-${data.id}` : ''
+    
+        // Lift any col- and row- type class to the form-group wrapper. The row- class denotes the row group it should go to
+        if(data.className){
+          const classes = data.className.split(' ').filter(className => (className.startsWith('col-') || className.startsWith('row-')));
+          if(classes && classes.length > 0){
+            className += ` ${classes.join(' ')}`;
+
+            // Now that the col- types were lifted, remove from the actual input field
```
kevinchappell: I'm not sure if this is working correctly.
image

src/js/form-render.js
```diff
@@ -82,10 +82,34 @@ class FormRender {
...
+            if(!rowGroupNode){              
+              rowGroupNode = document.createElement('div');
+              rowGroupNode.id = rowID;
+              rowGroupNode.classList.add('row');   
+              rowGroupNode.classList.add('form-inline');
+              renderedFormWrap.appendChild(rowGroupNode)
+            }
+            rowGroupNode.appendChild(field);
```
kevinchappell: nice :+1:

src/js/form-render.js
```diff
@@ -82,10 +82,34 @@ class FormRender {
...
+          if(classes[1]){           
+            const rowGroup = classes[1].split(' ')[0];
+            const rowID = this.id ? `${this.id}-row-${rowGroup}` : `row-${rowGroup}`;
+
+            // Check if this rowID is created yet or not.            
+            let rowGroupNode = document.getElementById(rowID);
+            if(!rowGroupNode){              
+              rowGroupNode = document.createElement('div');
+              rowGroupNode.id = rowID;
+              rowGroupNode.classList.add('row');
```
kevinchappell: do we need a prefix to avoid conflict? `fb-row`?

src/js/form-render.js
```diff
@@ -82,10 +82,34 @@ class FormRender {
           className: 'rendered-form',
         })
         this.appendChild(renderedFormWrap)
+  
         fields.forEach(field => {
-          renderedFormWrap.appendChild(field)
-          field.dispatchEvent(events.fieldRendered)
+          // Determine if rows are being used. If so, create the row and append to its row-{group}
+          // If the fields have row-, create & append to the appropriate row
+          const classes = field.className.split('row-');
```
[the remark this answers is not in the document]
Jojoshua: Putting a row-{number} was only for example. You might want to name the row to something more meaningful like row-employeeInfo. Then again, maybe the row- identifier really doesn't matter at all.  It really depends how it is implemented. For example if the row & columns could be configured via a drag & drop user interface then perhaps the row- identifier could be auto-generated. 

For now it would be best to have the regex match against row-{word} up until a space. So it might look like row-([^\s]+)

src/js/form-render.js
Jojoshua: Can't immediately apply the markup because as this keeps looping I need to check if the row is created or not for each field found. I can use utils to create the markup once I find it is not created.

src/js/layout.js
Jojoshua: So this implementation actually only applies to formRender not formBuilder. I did not know how to make it render with columns on the builder page, only once rendered. There is a disconnect in that you need to know what you applying without seeing the result immediately in the builder. I would toggle to see the end result. I am not familiar enough with builder to figure out how to make this work there as well.

src/js/form-render.js
Jojoshua: I don't think so because 'row' is the actual bootstrap class.

https://getbootstrap.com/docs/3.3/css/#grid

Jojoshua:
So yeah I did not know how to approach this with actually showing the preview inside formBuilder itself. 

A couple more ideas that could be done...

- Add a new config flag to use bs-columns. This flag would then automatically apply rows so that every time you added a form field, it went into the last row. It would then force you to drag the element down into a different row if you wanted it elsewhere. Also, could possibly auto set the col- classes via a resize drag....it all depends on how this should be targeted. For developers, adding the classes to the classList is natural and they understand what is happening. Having a separate property for the row- identifier would be appropriate.

Jojoshua:
Needed to add /^col-(xs|sm|md|lg)-([^\s]+) to match other bootstrap grid classes like col-md-push-2

kevinchappell:
I think this is ready to merge, in my testing I didn't see any issues. The majority of formBuilder users aren't developers and probably won't be adding columns this way but as long as the column functionality doesn't impact their experience there shouldn't be any issue. It would be great to eventually see formBuilder get to Formeo's level of column support but this is a great step in the right direction. 
columns

Jojoshua:
I hadn't seen formeo. Why did you make another project that does the same type of thing? The column support looks like a nice UI

kevinchappell:
> Why did you make another project that does the same type of thing?
at the time my thinking was that formBuilder needed so much work to get to a point of being able to support columns, conditions, repeatable input groups, control groups, etc. that i might as well start from scratch.

futureweb:
@Jojoshua @kevinchappell - trying this Feature with latest 3.6.1 Version. 
But never getting back the grouped Columns when requesting the HTML?

My Test FormData:
```
[
  {
    "type": "text",
    "required": false,
    "label": "Textfeld",
    "className": "form-control row-1 col-md-6",
    "name": "text-1599480302286",
    "subtype": "text"
  },
  {
    "type": "textarea",
    "required": false,
    "label": "Textbereich",
    "className": "form-control row-1 col-md-6",
    "name": "textarea-1599480303119"
  },
  {
    "type": "text",
    "required": false,
    "label": "Textfeld",
    "className": "form-control row-2 col-md-6",
    "name": "text-1599566007434",
    "subtype": "text"
  },
  {
    "type": "textarea",
    "required": false,
    "label": "Textbereich",
    "className": "form-control row-2 col-md-6",
    "name": "textarea-1599566005067"
  }
]
```

Rendering like:
```
renderedForm.formRender({formData});
formHTML = renderedForm.html();
console.log(renderedForm.html());
```

Output:
```
<div class="rendered-form">
   <div id="row-row-1" class="row form-inline">
      <div class="formbuilder-text form-group field-text-1599480302286 row-1 col-md-6"><label for="text-1599480302286" class="formbuilder-text-label">Textfeld</label><input type="text" class="form-control" name="text-1599480302286" id="text-1599480302286"></div>
   </div>
   <div id="row-row-1" class="row form-inline">
      <div class="formbuilder-textarea form-group field-textarea-1599480303119 row-1 col-md-6"><label for="textarea-1599480303119" class="formbuilder-textarea-label">Textbereich</label><textarea type="textarea" class="form-control" name="textarea-1599480303119" id="textarea-1599480303119"></textarea></div>
   </div>
   <div id="row-row-2" class="row form-inline">
      <div class="formbuilder-text form-group field-text-1599566007434 row-2 col-md-6"><label for="text-1599566007434" class="formbuilder-text-label">Textfeld</label><input type="text" class="form-control" name="text-1599566007434" id="text-1599566007434"></div>
   </div>
   <div id="row-row-2" class="row form-inline">
      <div class="formbuilder-textarea form-group field-textarea-1599566005067 row-2 col-md-6"><label for="textarea-1599566005067" class="formbuilder-textarea-label">Textbereich</label><textarea type="textarea" class="form-control" name="textarea-1599566005067" id="textarea-1599566005067"></textarea></div>
   </div>
</div>
```

As one can see - it's not grouping, but duplicating the IDs ... ?!? Hope you can give me a hint on what's wrong here as I would badly need this "Multi Column" Feature ;-)

Thank you, bye from Austria
Andreas

futureweb:
just figured out it's working when using an actual existing Dom element within the Page:
`var fbRender = document.getElementById("fombuilderrender");`
insted of a on the fly generated Element?!?
`var fbRender = $("<div>");`

Jojoshua:
> > Why did you make another project that does the same type of thing?
> 
> at the time my thinking was that formBuilder needed so much work to get to a point of being able to support columns, conditions, repeatable input groups, control groups, etc. that i might as well start from scratch.

Reflecting on this again, I think formBuilder codebase is still workable and can support all that but just needs some restructuring to make it feel extensible/clean again. Also, moving the codebase to Typescript would be a huge first step forward.

kevinchappell:
@Jojoshua a couple other reasons were themeability and not requiring jQuery. The number of changes required to support all of those features seemed like too much for me to handle alone, but with the support of you and the community I think its all doable. I agree there is still a need for formBuilder to continue to grow and moving to typescript would be great.

Jojoshua:
> @Jojoshua a couple other reasons were themeability and not requiring jQuery. The number of changes required to support all of those features seemed like too much for me to handle alone, but with the support of you and the community I think its all doable. I agree there is still a need for formBuilder to continue to grow and moving to typescript would be great.

Just FYI @kevinchappell I was playing around and did a TypeScript PoC. Not exactly sure how to do the webpack properly to build but got rid of all TS errors except for 8 on the formbuilder instance stuff which not really sure how to resolve. https://github.com/Jojoshua/formBuilder/tree/TypeScriptMigrationPoC

Obaidraj:
issue1234

PR #908, opened 2019-02-10
Define new methods object for every new formBuilder instance

Tarek-Essam:
When creating Two formBuilder on the same page the one that it's promise resolve last will override on the the actions object inside the instance. So we will have two instance of the formBuilder using the same actions like **setData** which for example call **removeAllFields** function using **d.stage** variable accessed by the closure. The result is having multiple formBuilders manipulating the same stage. 
I moved the definition of the methods objects inside **jQuery.fn.formBuilder** function so every time i create a new formBuilder i create a new methods object which solved the problem for me.

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1378,51 +1378,51 @@ const FormBuilder = function(opts, element) {
   return formBuilder
 }
 ;(function($) {
-  const methods = {
-    init: (options, elems) => {
-      const { i18n, ...opts } = jQuery.extend({}, defaultOptions, options, true)
-      config.opts = opts
-      const i18nOpts = jQuery.extend({}, defaultI18n, i18n, true)
-      methods.instance = {
-        actions: {
-          getFieldTypes: null,
-          addField: null,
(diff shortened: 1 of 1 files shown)


kevinchappell:
Thanks for the contribution!

Tarek-Essam:
You are welcome. Happy to help :)

mrityunjay38:
> When creating Two formBuilder on the same page the one that it's promise resolve last will override on the the actions object inside the instance. So we will have two instance of the formBuilder using the same actions like **setData** which for example call **removeAllFields** function using **d.stage** variable accessed by the closure. The result is having multiple formBuilders manipulating the same stage.
> I moved the definition of the methods objects inside **jQuery.fn.formBuilder** function so every time i create a new formBuilder i create a new methods object which solved the problem for me.

I am still facing this issue. When I create two formBuilders on the same page, the one which is created recently, overwrites functionality of the one which was created first.

Please, tell me how to create multiple formBuilders on the same page and each should operate seperately without interfareing each others  onSave() functionality.

Issue #916, opened 2019-03-01
onAddField 


Mordef:
### Description:

onAddField not fired

### Environment Details:

 - formBuilder Version: 3.1.3 
 - Browser: Firefox 65.0.2
 - OS: Win10

Taken example from https://formbuilder.online/docs/formBuilder/options/onAddField/

Mordef:
Just a mistake in the docs:
Headline: onAddField <-- works
Example: onFieldAdd <-- wrong

kevinchappell:
resolved in #936

asangster:
@kevinchappell Please update code sample in docs: https://formbuilder.readthedocs.io/en/latest/formBuilder/options/onAddField/

PR #936, opened 2019-04-23
Changed function name

jasonjewell:
The previous "onFieldAdd" option did not correctly call the function "onAddField". Update to function example to use correct call.


jasonjewell:
Hope this helps! It had me confused for a good while.

kevinchappell:
Excellent. This should resolve #916 

Thank you!

PR #941, opened 2019-05-01
Fix markdown for list elements (docs display)

BintLopez:
Fixes key files and folders list in the docs https://formbuilder.online/docs/formBuilder/overview/ ...
image

... to display as a list instead of a paragraph
image


kevinchappell:
excellent. this is much more readable. Thank you!

PR #950, opened 2019-05-20
Allow the textarea.quill control to be configured by passing in options

LiamKeene:
Using Quill in a project outside of formbuilder can causes issues if the versions are incompatible.

For example the 1.3.4 series will emit events from the Quill container differently to the 1.2.4 series (currently in the textarea.quill.js control), but both bind listeners to the same container.  

Added in a way to override the JS/CSS assets that are loaded for Quill and also for customising the Quill instance options.


kevinchappell:
Excelent work. Will test and likely merge tomorrow

Issue #999, opened 2019-10-17
FormBuilder: Max, min and step in formData does not work for number fields


nahojd:
When setting formData in formbuilder options, min, max and step for number fields does not work. They are not set in the UI for designing the form, and they are not present when the formdata is generated after save. What's more, if I set the value for a number field, it sets max, min and step as well.

Example options that does not work: 

```javascript
options = {
  formData: 
    '[{"type":"number", "label":"Max and min does not work for number", "max":100, "min":10,"step":2},' +
    '{"type":"number", "label":"Value overwrites max min and step", "max":100, "min":10,"step":2,"value":"50"},' +
    '{"type":"text", "label":"Maxlength works for text", "maxlength":5}]'
    };
```

Example in codepen: https://codepen.io/nahojd/pen/YzzGXJV

### Environment Details:

 - formBuilder Version: 3.2.5
 - Browser: Firefox
 - OS: Windows

buttle:
Hi,
I was just about to open an issue for the same thing. :)

My observations.
Max, min, and step are in fact saved correctly (I can see that in the database). But.. when the form is edited again; max, min, and step are lost.
I'm pretty sure this is because their values are not rendered as input values (inputs are blank). That means that every time you edit a form, you have to remember to set those number-field attributes again before saving.

Do you find that too @nahojd ?

nahojd:
@buttle Yes, that's my experience as well. When the max, min or step attributes are set, they are saved correctly. But when you load the saved data again, they are not set anymore.

stefansn:
@nahojd have you found a solution for this problem?

nahojd:
@stefansn Well, I have found the bug in the formbuilder, but it needs to be fixed in the code.

In `form-builder.js`, in the `numberAttribute` function starting on line 706, it actually removes the min, max and step attributes and passes them on as attributes instead of values, which sounds good, but is wrong. The result is that instead of the configuration input fields for min, max and step having the value of min, max and step respectively, instead they get the _actual min, max and step attributes_.

So the solution is to change this:

```javascript
 const numberAttribute = (attribute, values) => {
    const { class: classname, className, min = 0, max, step, value, ...attrs } = values
    const attrVal = attrs[attribute] || value
    const attrLabel = mi18n.get(attribute) || attribute
    const placeholder = mi18n.get(`placeholder.${attribute}`)

    const inputConfig = {
      type: 'number',
      value: attrVal,
      name: attribute,
      min,
      max,
      step,
[...]

PR #1018, opened 2019-12-30
Fix for bug in numberAttribute method where min, max and step are passed on as attributes instead of as value

nahojd:
This is a fix for the issue in https://github.com/kevinchappell/formBuilder/issues/999


kevinchappell:
min, max and step are intentionally attributes https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#max

Can you share a codesandbox or jsfiddle of the expected result?

kevinchappell:
looks good. 👍

Issue #1022, opened 2020-01-03
form builder and form render not working with internet explorer 11


harshalyelpale:
### Description: 
App Env. : Angular 8 with .Net Core 3.0
form builder and form render not working with internet explorer 11, even edge browser also.
Tried with polyfill.js also, below is script added at the top of other JS files

`<script src="https://polyfill.io/v3/polyfill.min.js?features=Object.assign,Object.entries,Promise,Array.prototype.includes,Array.prototype.find,String.prototype.startsWith,Object.values,String.prototype.includes"></script>`

![formBuilderPolyfillUse](https://user-images.githubusercontent.com/29271510/71709906-aed0ba00-2e1f-11ea-810b-d07fe3275d4c.png)

### Environment Details:

 - formBuilder Version: 
 - Browser: IE 11 & Edge
 - OS: Windows 7, Windows 10

W1-PopelierE:
This is the error:

`{description: "Unspecified error.", 
message: "Unspecified error.", number: -2147467259, stack: "Error: Unspecified error. at t.orderFields (https://formbuilder.online/assets/js/form-builder.min.js:19:102944) at t.appendControls (https://formbuilder.online/assets/js/form-builder.min.js:19:103526) at t.init (https://formbuilder.online/assets/js/form-builder.min.js:19:101748) at e (https://formbuilder.online/assets/js/form-builder.min.js:19:101543) at A (https://formbuilder.online/assets/js/form-builder.min.js:19:104228) at Anonymous function (https://formbuilder.online/assets/js/form-builder.min.js:19:126310) at each (https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.4/jquery.min.js:2:2875) at n.prototype.each (https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.4/jquery.min.js:2:838) at Anonymous function (https://formbuilder.online/assets/js/form-builder.min.js:19:126290)"}`


When running the project in a node project it works in Edge (https://formbuilder.online/docs/contributing/).

kevinchappell:
can you share your configuration in a codepen?
@W1-PopelierE can you test #1027 in IE11?

W1-PopelierE:
> can you share your configuration in a codepen?
> @W1-PopelierE can you test #1027 in IE11?

Will test this when I get home from work.

W1-PopelierE:
@kevinchappell in IE11 i get these errors:
[...]

PR #1033, opened 2020-01-31
Fixed "Invalid calling object" error caused by "console.log" in internet explorer.

harshar1987:
### Description:
formRender fails with error "Invalid calling object" in Internet Explorer 11. Due to this we get the return value as "undefined" from the "formRender" function. 

### Environment Details:

 - formBuilder Version: 3.3.1 (Used with ReactJS)
 - Browser: Internet Explorer 11
 - OS: Windows 10

### Steps to Reproduce
The issue is with the way console's functions are used in form-render.js and config.js.
image

This approach of assinging "console.log" / "console.warn" to a variable and invoking that variable as function does not work in internet explorer.

Below is sample code to reproduce. 
```HTML 
<!DOCTYPE html>
<html>
<head>
<title>Page Title</title>
</head>
<body>
<script>
alert(console.log);  // shows "native function"

var log = console.log;
alert(log);      // shows "native function"

console.log("You ain't gonna see me"); // silently ignored

log("I'm gonna crash!"); // result in an exception

alert("No exception");
</script> 
</body>
</html>
```

If we run the above code in internet explorer then the alert shows the native function". Notice that the alert message "No exception" isn't shown.

image

If "Script debugging is enabled" then we can see the error "Invalid Calling Object"

image

If we open the "Developer Tools" and rerun this code than it works. Notice that the alert now shows a different function code block with a "try"/"catch" and also we can see alert message "No exception".

image

image

Also before IE 10, the "console" object isn't available by default and only available if the developer tools are open. If someone needs support for older browser than IE 10 then a polyfill like below may need to be added.
```Javascript
if(!window.console) {
  var console = {
    log : function(){},
    warn : function(){},
    error : function(){},
  }
}
```
I am bit unsure, if changing this would have any impact. Please review and let me know your thoughts or suggestions or if any modifications are needed.

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -41,9 +41,15 @@ export const defaultOptions = {
   fieldEditContainer: null, // DOM node or selector
   inputSets: [], // add groups of fields at a time
   notify: {
-    error: console.error,
-    success: console.log,
-    warning: console.warn,
+    error: error => {
+      console.log(error)
+    },
+    success: success => {
+      console.log(success)
+    },
+    warning: warning => {
+      console.warn(warning)
+    },
   },
   onAddField: (fieldId, fieldData) => fieldData,
   onClearAll: noop,
--- a/src/js/form-render.js
+++ b/src/js/form-render.js
@@ -240,7 +240,7 @@ class FormRender {
     const fieldData = opts.formData
     if (!fieldData || Array.isArray(fieldData)) {
       throw new Error(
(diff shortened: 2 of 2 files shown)


W1-PopelierE:
But the big question now, is the form builder working in ie11 after this fix? Or is it still not working?

harshar1987:
Ah yes, it works perfectly fine after this fix :)

kevinchappell:
Resolves #1022

Issue #1139, opened 2020-11-25
Expected linebreaks to be 'LF' but found 'CRLF'


chandler3224:
### Description:
Since the addition of .eslintrc.json, whenever a build is triggered on my Windows machine I receive the following error for multiple files:
Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style

Removing "linebreak-style": ["error", "unix"], lets the process work. 

### Environment Details:

 - OS: Windows 10

### Expected Behavior
Successful Build

### Actual Behavior
Failed build.

### Steps to Reproduce
Run npm run-script build

chandler3224:
@kevinchappell May be worth adding a .gitattributes file with 
`* text eol=lf`
However I think some files may need to be normalised.

Issue #1049, opened 2020-03-11
Docs not clear on defining attrs for custom fields 


mfreeman-xtivia:
I have read and then re-read the FormBuilder docs and for the life of me its not obvious.

If i am building a new custom control Foo, and it needs to support user-editable attributes A, B, and C, how and where do I define that for my custom control?

chandler3224:
There is a new option named defaultAttrs. Docs updated in docs/formBuilder/overview.md

PR #1140, opened 2020-11-26
Default attributes in control plugins

chandler3224:
Adds the ability to define attributes for control plugins without needing to update the formBuilder options during init. 
This also solves the following items:

#1139 
#1049


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
kevinchappell, line 1:
```suggestion
* text eol=crlf
```

kevinchappell (changes requested):
i really like this feature but can we have the `eol` setting be `crlf`?

.gitattributes
chandler3224: Can't see why not, though it may flag up changes on a bunch of files as these were all previously ```lf```.

kevinchappell (approved):
lgtm! thank you!

.gitattributes
kevinchappell: i think the `lf` setting is what caused these non-change changes in your pr. 
image


not a big deal but it now looks like you've made tons of changes. so many that you made the top contributors on the website :sweat_smile: 

image

chandler3224: Ah :/ 
That's slightly embarrassing. I only made the change as it was breaking the build process and the majority of the other files were LF anyway. By all means do what you need to do to drop me from that list haha

kevinchappell: haha. thanks for the link it buts all good.

Issue #1046, opened 2020-03-05
typeUserAttrs only when specific SubType is selected?


futureweb:
Hey there,
is is possible to add a special typeUserAttrs only when "E-Mail" SubType in Textfield is selected?
I need a DropDown when E-Mail SubType is selected with "Empty Value", "Set as Sender".
thx, bye from Austria
Andreas

futureweb:
solved as own subtype ...

```
subtypes: {
            text: ["email_als_Absender"]
         }
```

parrainc:
Any updates about this? 
I'm needing the same thing, but looks like that's no possible yet :(

Have you figured out how to do it?

parrainc:
I've managed to implement this, so far it's working as I needed. 
@futureweb I think this could work for your use case as long as you set the subtype to the text field before rendering. (not sure if the library would support something like what you mentioned though which would be adding the field dynamically based on the subtype).

I can create a pull request, probably the way I implemented it is useful in some way to get this merged.

futureweb:
@parrainc - for our use case it would unfortunately have to happen dynamically after rendering.
Our customers can choose in the text field whether it is normal "text", "password", "e-mail" or a few other types. And depending on the selected type, you should be able to select a few more attributes. For example, in the case of "E-Mail", the attribute "must be empty" (for anti-spam checks) ...
But I guess your PW can help others ... ;-)

parrainc:
I hope so, and also would like to see that feature you need implemented soon. There might be some users who benefits from it.
cheers!

lucasnetau:
Was implemented in PR https://github.com/kevinchappell/formBuilder/pull/1168

PR #1168, opened 2021-03-14
Support the ability to define user attributes for subtype

parrainc:
This one tries to solve one part of #1046 (the ability to create user attributes based on subtypes)
Currently, only user attributes based on types are able to be added to the formBuilder instantiation process.

Please, review and let me know if it looks good or needs any adjustments.
(already working fine in my project)

### DIFF
--- a/src/demo/js/demo.js
+++ b/src/demo/js/demo.js
@@ -56,6 +56,16 @@ jQuery(function($) {
       },
       icon: '🌟',
     },
+    {
+      type: 'checkbox-group',
+      subtype: 'custom-group',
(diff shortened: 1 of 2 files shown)


parrainc:
I just figured that once you saved the form and loaded again, the fields get lost from the ui. That's because the checkbox group does not have the Type dropdown with the subtype. I'm looking to see if that's possible to add it, that way the declared subtype is still in the element once saved/loaded.

kevinchappell (approved):
wicked. this feature was long overdue

PR #1199, opened 2021-06-01
Add a per epoch counter to generate unique names

### LINKED ISSUE #1198: nameAttr can return conflicting names when adding multiple elements of the same type
### Description:
nameAttr in utils.js can return the identical name for two elements created at the same time. Firefox exhibits this issue when creating a form from a template without any names associated with elements.

prefix + '-' + epoch is not unique enough

### Environment Details:

 - formBuilder Version: 3.7
 - Browser: Firefox (Easy to reproduce on)
 - OS: Any

### Expected Behavior
Name returned by nameAttr is unique

### Actual Behavior
Name returned by nameAttr can conflict

### Steps to Reproduce
Initialise a formBuilder with formData set with multiple fields of the same time without a name attribute

lucasnetau:
Fixes #1198 this is a proposed fix to ensure unique names are generated names for identical field types at the same epoch. A per epoch counter is implemented.

Changes: New elements now have a default value of prefix-epoch-counter starting from 0

I also investigated using performance.now(), however on Safari and Firefox the resolution for performance.now() is no greater than Date().getTime().

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -162,11 +162,21 @@ export const bindEvents = (element, events) => {
  * @param  {Object} field
  * @return {String}       name
  */
-export const nameAttr = function(field) {
-  const epoch = new Date().getTime()
-  const prefix = field.type || hyphenCase(field.label)
-  return prefix + '-' + epoch
-}
+export const nameAttr = (function() {
+  let lepoch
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
nice change!

Issue #1152, opened 2021-01-01
Multiple in select option: BUG


nguyendachuy:
### Description:
I use the normal builder select but when rendering the HTML it automatically converts name select to array

### Environment Details:

 - formBuilder Version: 3.6.1
 - Browser: Firefox 85.0b4 (64-bit)
 - OS: Windows 10

### Screenshot - *(optional)*
![image](https://user-images.githubusercontent.com/15666602/103434557-b3e46700-4c35-11eb-9bbb-e79472409cd9.png)
![image](https://user-images.githubusercontent.com/15666602/103434563-d2e2f900-4c35-11eb-9536-8d0959647b70.png)

KrenZfer:
Hi,
I use this to fix the that like this,

`// fixAttr
$("[multiple=false]").removeAttr("multiple");
$("[readonly=false]").removeAttr("readonly");`

actually that because our web browser (not all maybe) interpret all select that have multiple attribut even it have false value as multiple select so we have to eliminate that attribute if it has attribute if its value is false,

Let me know if its work for you or not

nguyendachuy:
Your fix can be fixed. But I went back to an older version. Thanks

PR #1216, opened 2021-08-17
Ensure boolean attributes are set only when true and to a valid value

### LINKED ISSUE #1209: [Bug] Fileuploads (input type="file") always multiple
### Description: checkbox "Allow users to upload multiple files" does not work any more

if checked, it creates the attribute multiple="true"
if unchecked, it creates the attribute multiple="false"

in html similar to the attributes required, disabled and so on, the value is not used at all

just being there, results in always TRUE
'multiple' == 'multiple="multiple"' == 'multiple="true"' == 'multiple="false"'

<input type="file" class="form-control" name="file-1571296005358" **multiple="false"** id="file-1571296005358">
^-- this is a an input field, that allows multiple uploads

### Environment Details:

 - formBuilder Version: 3.6.1
 - Browser: Firefox
 - OS: Windows

### Expected Behavior

no 'multiple' attribute at all in case of the checkbox  "Allow users to upload multiple files" being unchecked

### Actual Behavior

see description

### Steps to Reproduce

see description

lucasnetau:
Currently, in utils.js and select.js boolean attributes are set to "true" and "false" which are not valid under the HTML5 specification https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes / https://www.w3.org/TR/html52/infrastructure.html#sec-boolean-attributes

Per the specification "If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute’s canonical name, with no leading or trailing white space."

Fixes #1209, #1152

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -197,7 +197,7 @@ export default class controlSelect extends control {
 
           for (let i = 0; i < selectedOptions.length; i++) {
             if (input.value === selectedOptions[i]) {
-              input.setAttribute('checked', true)
+              input.setAttribute('checked', 'checked')
               selectedOptions.splice(i, 1) // Remove this item from the list
               break
             }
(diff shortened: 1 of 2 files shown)


lucasnetau:
@kevinchappell Can you please merge this PR? The current code breaks forms where a false flag is wanted.

kevinchappell (approved):
lgtm!

PR #1219, opened 2021-08-24
Remove type="textarea" from textarea elements to match HTML5 spec

lucasnetau:
Text areas rendered by FormBuilder fail the W3C https://validator.w3.org/ due to element containing the attribute 'type' set to 'textarea'. This patch drops the type attribute in the control's build() step before passing the data to utils.markup for rendering

### DIFF
--- a/src/js/control/textarea.js
+++ b/src/js/control/textarea.js
@@ -23,6 +23,8 @@ export default class controlTextarea extends control {
    */
   build() {
     const { value = '', ...attrs } = this.config
+    //Textareas do not have an attribute 'type'
+    delete attrs['type']
     this.field = this.markup('textarea', this.parsedHtml(value), attrs)
     return this.field
   }
--- a/src/js/control/textarea.quill.js
+++ b/src/js/control/textarea.quill.js
@@ -45,6 +45,8 @@ export default class controlQuill extends controlTextarea {
   build() {
     // eslint-disable-next-line no-unused-vars
     const { value = '', ...attrs } = this.config
+    //Textareas do not have an attribute 'type'
+    delete attrs['type']
     this.field = this.markup('div', null, attrs)
     return this.field
   }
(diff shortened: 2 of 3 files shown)


src/js/control/textarea.js
```diff
@@ -23,6 +23,8 @@ export default class controlTextarea extends control {
    */
   build() {
     const { value = '', ...attrs } = this.config
+    //Textareas do not have an attribute 'type'
+    delete attrs['type']
```
kevinchappell, line 27: We might not want a `delete` here because it will remove it completely from the scope and other parts of the project may require the `type` attribute as its used a lot. I think safer would be to destructure it out and not use it.
```suggestion
    const { value = '', type: _type, ...attrs } = this.config
    // Textareas do not have an attribute 'type'
```

kevinchappell (approved):
sorry so late in reviewing this PR. I just had one comment about destructuring out the `type` instead of deleting it but seeing how this PR is getting stale I'm happy to approve now and address the delete if it becomes an issue later. Will wait a couple days in case you want to make that change then merge.

Thanks @lucasnetau !

src/js/control/textarea.js
lucasnetau: I thought about doing that, however the linter throws an no-unused-vars warning

```
WARNING in ../src/js/control/textarea.js
Module Warning (from ../node_modules/eslint-loader/dist/cjs.js):

../src/js/control/textarea.js
  26:31  warning  '_type' is assigned a value but never used  no-unused-vars

✖ 1 problem (0 errors, 1 warning)
```

I couldn't see where deleting type from the attrs array will have any downstream effect. The element's config property still contains the type value. attrs is only used for building the textarea element via this.markup

kevinchappell: ok lets leave it as-is for now then. We should add the following rule to the linter config:

```javascript
  'unused-imports/no-unused-vars': [
      'warn',
      {
        argsIgnorePattern: '^_',
        varsIgnorePattern: '^_',
      },
    ],
```

PR #1223, opened 2021-09-16
Adding example to onAddOption.md

jamiejones85:
Adding an example usage, as it wasn't clear this would do what I wanted to achieve.


kevinchappell (approved):
awesome!

PR #1233, opened 2021-11-09
Fix rendering of required checkbox-groups with userData

### LINKED ISSUE #1218: CHECKBOX: if required = true, then checkbox items remains unselected
### Description:
For the checkbox-group, if **required = true**, then checkbox items remains unselected even if userData provided.

You can check here, https://codepen.io/mvir/pen/wvdVmQd

### Environment Details:

 - formBuilder Version: 3.7.3
 - Browser: Google Chrome
 - OS: Windows 10

### Expected Behavior
formRender-userData1

### Actual Behavior
formRender-userData2

### Steps to Reproduce
https://codepen.io/mvir/pen/wvdVmQd

lucasnetau:
Override of onRender breaks checkbox-groups which have the required flag set along with provided userData. Ensure we call both the groupRequired function along with the original onRender.

Fixes #1218

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -33,7 +33,12 @@ export default class controlSelect extends control {
     }
 
     if (type === 'checkbox-group' && data.required) {
-      this.onRender = this.groupRequired
+      const self = this
+      const defaultOnRender = this.onRender.bind(this)
+      this.onRender = function() {
+        self.groupRequired()
+        defaultOnRender()
+      }
     }
 
     delete data.title


lucasnetau:
@kevinchappell Can you please check and merge this PR? There are open bug tickets that it resolves.

Lukimer (commented):
Dear lucasnetau, you are my angle, you save my life, but I am just able to use the formRender.min.js only, and how can I compile this changes of select.js and rebuild to the formRender.min.js??  or could you kindly teach me or send me the updated formRender.min.js??

Thank you very much!!!!!
Best Regards,
Kim

lucasnetau:
@Lukimer ,  follow the instructions here https://formbuilder.online/docs/development/ to download the repo and install the dependencies.

Instead of `npm start` you can run `npm run-script build` to just build the files

Then grab the formRender.min.js file from `dist/form-builder.min.js`

Lukimer:
@lucasnetau , heartly thanks for your fix and instructions for me, you are my God!!

And I tried to use your fix code directly modified to formrender.min.js, also worked. And actually I also found there is similar bug for radio group too. I also used your fix code for fixing radio group too. I share below help can help other friends:

For form-render.minjs:

 if ("checkbox-group" === i && c.required) {                        
                        const self = this
                        const defaultOnRender = this.onRender.bind(this)
                        this.onRender = function () {
                          defaultOnRender()
                          self.groupRequired()
                          defaultOnRender()
                        }
 if ("radio-group" === i && c.required) {                                            
                        const self = this
                        const defaultOnRender = this.onRender.bind(this)
                        this.onRender = function () {
                          defaultOnRender()
                          self.groupRequired()
                          defaultOnRender()
                        }
                    }

lucasnetau:
@Lukimer , can you explain why radio-group needs to be covered? groupRequired() has only ever been applied fro checkbox-group since it handles the custom validation for checkbox groups which can have 1 or more required responses. radio-groups don't have multiple responses so do not need the custom validator.

kevinchappell (approved):
Looks good. At some point these classes should be refactored to not be just functions so we can do something like 
```
this.onRender = () => {}
```

The arrow function would negate the need for a bind but this is fine for now.

Thank you!

PR #1248, opened 2022-01-26
Include subtype when using Copy

Jojoshua:
Copying a field with subtype doesn't clone the subtype properly. (tinymce, email, etc)


Obaidraj:
issue1234

PR #1277, opened 2022-03-19
Ensure groupRequire events run after applying any user data to checkbox groups

### LINKED ISSUE #1259: required checkboxes group
### Description:
in control - select.js 

    if (type === 'checkbox-group' && data.required) {
      this.onRender = this.groupRequired
    }

when "      if (type === 'checkbox-group' && data.required) {  " -- true 

 checkbox data never populated with saved data

lucasnetau:
groupRequired events should be attached after rendering the element. This is due to the fact that no change event is triggered after any checkboxes are checked via the onRender user data code.

Resolves #1259

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -36,8 +36,8 @@ export default class controlSelect extends control {
       const self = this
       const defaultOnRender = this.onRender.bind(this)
       this.onRender = function() {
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
nice! good fix

Issue #1295, opened 2022-04-01
FormBuilder getFormData() now triggers a closeAllEdit on the stage


lucasnetau:
### Description:

Since version 3.8.0 (https://github.com/kevinchappell/formBuilder/commit/c564e3515f0b6fdb7b65be1e334140620d2cc71d) getFormData() issues a closeAllEdit. If events are bound to formBuilder contenteditable fields tracking mutations then the edit fields are closed on every key press (In my particular case it is sanatising html input into contenteditiable labels)

https://github.com/kevinchappell/formBuilder/blame/master/src/js/helpers.js#L1268

### Environment Details:

 - formBuilder Version: >= 3.8.0
 - Browser: Any
 - OS: Any

### Expected Behavior
When calling FormBuilder.getFormData() the method should only return data, it should not change anything on the editing stage

### Actual Behavior
Editing stage is closed

Jojoshua:
I believe the reason for this is that during edit the preview container is actually outside its parent row container so calling get data would have lost all the fields currently open in preview.

lucasnetau:
@Jojoshua , I've never had any dataloss for over 12 months calling getFormData with the stage open. I use this also to track changes in the form and know when to prompt the user to save unsaved changes. The preview container isn't what is saved, so I do not understand why this would be of concern.

The function is named getFormData and has a very straight forward description of what it does, calling getFormData should not cause any other mutations to the system.

Jojoshua:
This is a change with the enhanced grid mode feature. I don't remember how far reaching the feature goes but data would be lost if the field wasn't put back in its proper place in the dom. This was so that the preview could still open up as it did before and not have to open in a modal. You might be able to wrap your suggested change in a check only if not using the new enhanced grid mode.

lucasnetau:
@Jojoshua , perhaps as the author of the grid mode you can please look into an appropriate fix if you believe reverting the change will break the grid functionality. Especially since the grid functionality was meant to be behind a feature flag and not cause any disruption if not enabled. This change has caused a BC break and a function named getFormData should not be causing the stage to collapse.

Jojoshua:
@lucasnetau thanks for reminding me about this. I checked into it and the thing I was concerned about looks to be already implemented (https://github.com/kevinchappell/formBuilder/pull/1243/commits/870b7aa8c27c1060217a7ee9af1fd7a15d13a4f4#r855223484) so I am OK with your PR

PR #1296, opened 2022-04-02
Revert change to FormBuilder.getFormData which triggers the stage to close on call

lucasnetau:
Fixes https://github.com/kevinchappell/formBuilder/issues/1295

Version 3.8.0 added a closeAllEdit() call to getFormData() which causes the stage to close. getFormData should not cause any side effects on call, it should only return the formData


lucasnetau:
@kevinchappell, any feedback on this PR? Causing the stage to collapse when retrieving getFormData is a BC break and not expected from the function's signature

Issue #1061, opened 2020-04-07
Checkbox Group "Other" - missing "name" Attr


futureweb:
When enabling "Other" in Checkbox Groups - the Other Text-Field got no "name" Attribute - so it won't be sent along other Data in Post ... "name" Attribute should be added.

![grafik](https://user-images.githubusercontent.com/1698881/78650474-3fa3af80-78bf-11ea-91ed-2d6012b33e19.png)

sebastian-alfers:
@kevinchappell is there a release (`.min.js`) that contains this fix? Not sure how to get it from the plain source code. Thanks

sebastian-alfers:
Asking as when I `npm i formBuilder@3.6.2v` the file `form-builder.min.js` says `Version: 3.6.1` in the header comment

sebastian-alfers:
It also tried it with `3.4.4` (which somehow needs `npm i formBuilder@3.4.5` to be installed) and it still does not render without the name attribute.

Issue #483, opened 2017-05-22
Radio "other" not in same radio group


CaptainHypertext:
### Description:
When you have a radio group in your form with "Let users to enter an unlisted option" selected, the "Other" radio button can be selected while another radio button in the same group is still selected.

### Environment Details:
 - formBuilder Version: 
 - Browser: Chrome 58
 - OS: Windows 10

### Expected Behavior
The other radio buttons should be de-selected.

### Actual Behavior
One of the other radio buttons remains selected

### Steps to Reproduce
Start a form, add a radio group, turn on "Let users to enter an unlisted option", then go back and click the Other radio option.  The text box appears, but Option 1 is still selected.

### Screenshot - *(optional)*
![formbuilderradio](https://cloud.githubusercontent.com/assets/15058607/26327077/e5ef7194-3f0b-11e7-89e0-cd33a4d1435b.PNG)

Note the html:
```
<div class="radio">
	<label for="radio-group-1495484752993-preview-2">
		<input name="radio-group-1495484752993-preview" class="" id="radio-group-1495484752993-preview-2" value="option-3" type="radio">Option 3
	</label>
</div>
<div class="radio">
	<label for="radio-group-1495484752993-preview-other" class="checked">
		<input name="radio-group-1495484752993-preview[other]" class=" other-option" id="radio-group-1495484752993-preview-other" type="radio">Other
		<input type="text" name="radio-group-1495484752993-preview" id="radio-group-1495484752993-preview-other-value" class="other-val" style="display: inline-block;">
	</label>
</div>
```

One is named `radio-group-1495484752993-preview` and the other is `radio-group-1495484752993-preview[other]`.

CaptainHypertext:
Thanks for the fix 👍

PR #1331, opened 2022-10-07
Fix incorrect re-assignment of Other input's name attribute

### LINKED ISSUE #1172: Radio Button & Other Option not working
### Description:
When we use the Radio Button list and the "Other" option together we run into a validation error, because it is possible to mark two elements in the radio button list. Validation fails.
We would like to see this solved meanwhile we don't allow to use the "Other" Option.
Checkboxes are working fine. 

### Environment Details:

 - formBuilder Version: 3.6.0
 - Browser: Edge, Chrome
 - OS: Windows

### Actual Behavior
When we use the Radio Button list and the "Other" option together we run into a validation error, because it ist possible to to mark two elements in the radio button list. It seems to be two lists. When the "Other" radio button gets selected and aome text is asses to the text field, the name of the <input type=radio name=blabla...   this gets automatically added to the name: []
### Screenshot - *(optional)*
image
image
image

### LINKED ISSUE #1318: radio-group issues when checking Others
### Description:
When I check Others and type in the textbox, the name attribute changes.
So the check remains.
### Environment Details:

 - formBuilder Version: last version
 - Browser: 
 - OS: 

### Expected Behavior
Do not change the name attribute

### Screenshot - *(optional)*
ss

lucasnetau:
The input handler for entry into the Other text input field would override the primary input name with a multiple [] appended name. This breaks radio-groups because the other now has a different name and therefore you can select the other option along with any provided option.

I cannot see where this name change would be required. The primaryInput field (radio or checkbox) gets its name from data.name which is the correct value for checkbox-group and radio-group. Select and multi-select do not support other option. 

https://github.com/kevinchappell/formBuilder/blob/013d0f91bb281a01bfdf9a774621d764fd489b78/src/js/control/select.js#L117

The only place I haven't tested is the singular checkbox control, however that is a disabled control per select.js https://github.com/kevinchappell/formBuilder/blob/013d0f91bb281a01bfdf9a774621d764fd489b78/src/js/control/select.js#L15

Fixes #1318
Fixes #1172

@kevinchappell  This appeared to be changed a few times in #1061, #483, #485 to fix radio, then checkbox however after reverting https://github.com/kevinchappell/formBuilder/commit/7e7569ef770d2adc568b1663c5a5ebb4095cf2c9 both checkbox-group and radio-group are working correctly for me (tested Firefox and Chrome)

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -124,7 +124,6 @@ export default class controlSelect extends control {
               const otherInput = evt.target
               const other = otherInput.parentElement.previousElementSibling
               other.value = otherInput.value
-              other.name = `${data.id}[]`
             },
           },
           id: `${otherOptionAttrs.id}-value`,


lucasnetau:
@kevinchappell any chance these PRs can be merged?

jayden626:
Just ran into this issue, thanks for the fix @lucasnetau

lucasnetau:
@kevinchappell Is formBuilder still maintained?

kevinchappell (approved):
👍️

PR #1332, opened 2022-10-07
Incorrect class list generated for other- option if class is not set

lucasnetau:
Small fix. If not className is assigned to the control then the class attribute for the other option becomes "undefined other-option", null coalesce to an empty string.


kevinchappell (approved):
Good catch

PR #1334, opened 2022-10-08
Fix input touchstart event handling

lucasnetau:
The input element is contained in e.target, 'this' refers to $stage in the callback


kevinchappell (approved):
Nice!

PR #1338, opened 2022-10-26
Fix numberAttributes with a valid value of 0

### LINKED ISSUE #1337: It does not load the min/max 0 of the input number
### Description:
As I have indicated in the title, when I select the input number, and in the min/max I set it to 0, save the data, and reload the form (refreshing the page), the 0 of the min/max disappears, however , in the json the min/max appears at 0, the problem is when I save again, as the min/max was empty it is removed from the json

lucasnetau:
A number attribute's value may be any Number including zero, cannot check existence of the value using ||, instead check if the value is NaN

Fixes #1337


kevinchappell (approved):
Good fix

PR #1382, opened 2023-07-05
Calculate key before src variable is overwritten

lucasnetau:
src was overwritten and make a string in the line prior to calculating the key. 

If no id is set in src.id then the key is calculated incorrectly as undefined


kevinchappell (approved):
Good catch

PR #1384, opened 2023-07-28
Use the created element when setting userData value

lucasnetau:
The form container element may not be included in the DOM when we render the form. If this is the case then the textarea and hidden element will not be found by the jQuery selector and no user value will be set. Change brings these two controls into the same onRender functionality of other controls like text.js


kevinchappell (approved):
These are good changes

PR #1385, opened 2023-08-05
Fix translation of "Other" label in rendered radio/checkbox group

### LINKED ISSUE #1163: [i18n][override] - "Other" label
### Description:

I have trouble customising language in UI.
Changes (as override) in i18n setting does not appear in rendered form.

### Steps to Reproduce

- Add Checkbox / Radio and allow custom user option listing
- See that label for custom option is "Other"
- Override lang setting in settings
```
const builderOptions = {
        i18n: {
            override: {
                'en-US': {
                    other: 'Text I want to see',
...
 
```
- Observe, that label for "Other" remain the same as before overriding in option

### Expected Behavior
Should result in different (specified) label text for "Other" option in checkbox/radio 

### Actual Behavior
Other label is rendered as "Other" despite change in lang settings

### Environment Details:
 - formBuilder Version: 3.6.1
 - Browser: Chrome/Chromium
 - OS: macOS

### Possible Resolution 

What I have found is that there is

https://github.com/kevinchappell/formBuilder/blob/902206505760b8af8417f479a4ddcdc641c46b10/src/js/control/select.js#L129

with hard-coded "Other" string instead of referece to i18n.

If that's the case I would be more than happy to hear guidance on how to fix this and submit pull request.

### LINKED ISSUE #1228: Missing translation for snippet for "Other"
### Description:
We translated the snippet "Other":
```js
  i18n:
  {
    override: {
      'en-US': {
        other: "Sonstige"
      }
    }
  }
```

But in the formBuilder and formRender it still outputs 'Other' instead of 'Sonstige'. 

### Environment Details:

 - formBuilder Version: current, v3.7.3
 - Browser: all
 - OS: all

### Expected Behavior
Printing 'Sonstige' instead of 'Other'

### Actual Behavior
Printing 'Other' instead of translation

lucasnetau:
The "Other" option string is hard-coded to English.  Use the value from the language file.

Fixes #1163
Fixes #1228

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -130,7 +130,7 @@ export default class controlSelect extends control {
           className: 'other-val',
         }
         const primaryInput = this.markup('input', null, optionAttrs)
-        const otherInputs = [document.createTextNode('Other'), this.markup('input', null, otherValAttrs)]
+        const otherInputs = [document.createTextNode(control.mi18n('other')), this.markup('input', null, otherValAttrs)]
         const inputLabel = this.markup('label', otherInputs, { for: optionAttrs.id })
         const wrapper = this.markup('div', [primaryInput, inputLabel], { className: wrapperClass })
         options.push(wrapper)


kevinchappell (approved):
Perfect

PR #1392, opened 2023-08-21
Use the i18n translated label for boolean typeUserAttrs. 

### LINKED ISSUE #1391: typeUserAttrs checkbox label does not use i18n value
### Description:
When adding a readonly checkbox attribute to a text field the label does not use the value specified in the i18n structure whereas the name attribute of a text field will use the i18n value.

### Environment Details:

 - formBuilder Version: 3.8.2
 - Browser: Chrome 115.0.5790.171 (Official Build) (64-bit)
 - OS: Windows 10

### Expected Behavior
A typeUserAttrs checkbox sub-type should retrieve it's label from the i18n structure.

### Actual Behavior
A typeUserAttrs checkbox sub-type retrieves it's label from the i18n structure.

### Steps to Reproduce
Single page repro.

Drag a text field onto the form builder
Edit the component
The name attribute is renamed to Field Id from the i18n structure
The Read-only field takes its label from the attribute name instead of the i18n structure.

See this code pen https://codepen.io/Matthew-Pritchard/pen/KKbPbxQ for a repro of the problem (adapted from the original  i18n sample https://codepen.io/kevinchappell/pen/EMqvbG)

lucasnetau:
Boolean userTypeAttrs were using the label directly instead of using the translated value

Fixes #1391

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -686,7 +686,7 @@ function FormBuilder(opts, element, $) {
         } else if (attrData.hasOwnProperty('value') || attrData.hasOwnProperty('checked')) {
           isChecked = attrData.value || attrData.checked || false
         }
-        return boolAttribute(attr, { ...attrData, [attr]: isChecked }, { first: attrData.label })
+        return boolAttribute(attr, { ...attrData, [attr]: isChecked }, { first: i18n[attr] })
       },
     }
 


src/js/form-builder.js
```diff
@@ -686,7 +686,7 @@ function FormBuilder(opts, element, $) {
         } else if (attrData.hasOwnProperty('value') || attrData.hasOwnProperty('checked')) {
           isChecked = attrData.value || attrData.checked || false
         }
-        return boolAttribute(attr, { ...attrData, [attr]: isChecked }, { first: attrData.label })
```
kevinchappell, line 689: doesn't this override the `attrData.label` configuration?

src/js/form-builder.js
```diff
@@ -686,7 +686,7 @@ function FormBuilder(opts, element, $) {
...
+        return boolAttribute(attr, { ...attrData, [attr]: isChecked }, { first: i18n[attr] })
```
kevinchappell, line 689: would something like this work?
```suggestion
        return boolAttribute(attr, { ...attrData, [attr]: isChecked }, { first: attrData.label || i18n[attr] })
```

kevinchappell (approved):

src/js/form-builder.js
lucasnetau: I don't believe so. We calculate the label value on line 704 using either the translated value of falling back onto raw .label value.
https://github.com/kevinchappell/formBuilder/blob/27abe220d4dabe8d08a2ab64c28d8e98a986854f/src/js/form-builder.js#L704

This change mirrors how string and select attributes calculate the label's value

String and Select attributes use the i18n value directly:
https://github.com/kevinchappell/formBuilder/blob/27abe220d4dabe8d08a2ab64c28d8e98a986854f/src/js/form-builder.js#L740

https://github.com/kevinchappell/formBuilder/blob/27abe220d4dabe8d08a2ab64c28d8e98a986854f/src/js/form-builder.js#L794

Number attribute does it slightly different:
https://github.com/kevinchappell/formBuilder/blob/27abe220d4dabe8d08a2ab64c28d8e98a986854f/src/js/form-builder.js#L888

src/js/form-builder.js
lucasnetau: i18n[attr] is calculated from attrData.label.

The following Jest testcase fails if you use attrData.label in

```javascript
describe('FormBuilder stage names translated', () => {
  test('boolAttribute name translated', async () => {
    const config = {
      typeUserAttrs: {
        text: {
          readonly: {
            label: ["readOnly"],
            value: false,
            type: 'checkbox',
          }
        }
      },
      i18n: {
        override: {
          "en-US": {
            readOnly: "Is Readonly",
          }
        }
      }
    }

    const fbWrap = $('<div>')
    const fb = await $(fbWrap).formBuilder(config).promise
    const field = {
      type: 'text',
      class: 'form-control'
    }
    fb.actions.addField(field)
    expect(fbWrap.find('.readonly-wrap label').text()).toBe(config.i18n.override['en-US'].readOnly)
  })
})
```

With the suggested change the test case fails because the label is untranslated:

```
    Expected: "Is Readonly"
    Received: "readOnly"
```

kevinchappell: i see, thanks for explaining

Issue #1164, opened 2021-02-24
Not getting desired outcome with checkbox values.


jringeisen:
### Description:
I'm currently using this formBuilder in a production application and my users are having a couple of issues but the main one that I am here to address is if a users builds a form and adds a checkbox with 4 options and sets them to be checked by default so that their client can uncheck the boxes they would like, then when saved all the boxes returned checked instead of the ones that were unchecked. Any ideas?

### Environment Details:

 - formBuilder Version: Latest
 - Browser: Chrome
 - OS: Mac

### Expected Behavior
If checkboxes are checked by default then when a user unchecks and saves the rendered outcome should be unchecked boxes.

### Actual Behavior
The user is unchecking boxing, clicking save, and the boxes are rendering checked.

### Steps to Reproduce
Drag and drop the checkbox field. Add a couple of options and check the box to check the options by default.

### Screenshot - *(optional)*
I unchecked option 1 and option 2 and hit save, the rendered outcome is below:
<img width="574" alt="Screen Shot 2021-02-23 at 9 00 13 PM" src="https://user-images.githubusercontent.com/7864804/108934031-29f3c180-761a-11eb-9d36-cf29c362d618.png">

Wujek280:
Reproduced.
What's more checkboxes value doesn't change if `required` option is set. Can you relate to this bug as well @jringeisen ?

jringeisen:
@Wujek280 yes, that is also an issue. Values don’t save if field is required.

Wujek280:
What I've noticed is that value of checkbox get saved to userData but could not retrieve checkbox value @ render

I need form to be re-rendered.

So I came up with backend workaround.

Have to process each formData element and set default selection based on userData entry

`json
[{
    "name": "checkbox-group-1615303428460",
    "type": "checkbox-group",
    "label": "Checkboxes",
    "other": false,
    "inline": false,
    "toggle": false,
    "values": [{
        "label": "A",
        "value": "A",
        "selected": false <-- (workaround) need to process saved data to set "true" here
    }, {
        "label": "B",
        "value": "B",
        "selected": false
    }],
    "required": true,
    "userData": ["A"] <-- this is user input
}]

`

PR #1396, opened 2023-08-23
Fix loading of checkbox-group user data when default select state set in formData

lucasnetau:
Ensure that default selected checkboxes are unselected when loading userData that doesn't contain them.

Fixes https://github.com/kevinchappell/formBuilder/issues/1164


kevinchappell (approved):
nice :+1:

PR #1398, opened 2023-08-23
Fix incorrectly assigning a value to numberAttributes

lucasnetau:
fix: If no value was set for a numberAttribute the value attribute's)value was used instead. This causes min/max/step/rows/maxLen number attributes to incorrectly be assigned a value if they are not set in formData but a number exists in the value attribute

Fixes: https://github.com/kevinchappell/formBuilder/issues/999


lucasnetau:
Before: 
image

lucasnetau:
After:
image

kevinchappell (approved):

kevinchappell:
good fix!

Issue #1217, opened 2021-08-18
removeField NOT working AT ALL


Ale-Digitale:
Hi,

why **removeField** doesn't work at all ? Are you aware of this ?

please try it here... click on the REMOVE FIELD red button... nothing happens:

https://formbuilder.online/docs/formBuilder/actions/removeField/

Jojoshua:
Can you post a fiddle to reproduce?

lucasnetau:
Hi @Ale-Digitale thank you for your report.

Calling removeField without passing a fieldID was broken in https://github.com/kevinchappell/formBuilder/commit/e0c0f2ea8f204da6e11d8de938e27ca6687d5588

The const field variable is declared prior to the fallback to select the last form field and therefore can not be updated.

PR #1399, opened 2023-08-23
Fix for fallback in removeField when no fieldID is provided (or fieldID is not found)

lucasnetau:
fix: revert change made in https://github.com/kevinchappell/formBuilder/commit/e0c0f2ea8f204da6e11d8de938e27ca6687d5588 which declared field as a const prior to the fallback to selecting the last field in the form. This change also prevents incorrect removing of the last field in the case where a fieldID is provided but which was not found in the form

removeField will remove the field given in fieldID, or if no fieldID passed to the function then it will remove the last field in the form.

Fixes: https://github.com/kevinchappell/formBuilder/issues/1217

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -1002,7 +1002,7 @@ export default class Helpers {
   }
 
   /**
-   * Remove a field from the stage
+   * Remove a given field from the stage or the last field if no fieldID is provided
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
good fix. i wasnt aware this stopped working

PR #1407, opened 2023-09-11
Fix generation of CSS rules for mobile devices

lucasnetau:
Ensure generation of CSS rules for .formbuilder-mobile place the selector at the .form-wrap.form-builder level

This class is appended to div.form-wrap.form-builder and therefore the selector needs to be appended to the parent selector and not a descendant.

Incorrect selector:
```css
.form-wrap.form-builder .formbuilder-mobile .form-actions{width:100%}
.form-wrap.form-builder .formbuilder-mobile .field-actions{opacity:1}
```

Corrected selector:
```css
.formbuilder-mobile.form-wrap.form-builder .form-actions{width:100%}
.formbuilder-mobile.form-wrap.form-builder .field-actions{opacity:1}
```

@kevinchappell , I considered pulling all the mobile styles out into their own file however this is the least LOC modification required and doesn't change the order of declarations

### DIFF
--- a/src/sass/_controls.scss
+++ b/src/sass/_controls.scss
@@ -129,7 +129,7 @@
   }
 }
 
-.formbuilder-mobile {
+@at-root #{selector-append(".formbuilder-mobile", &)} {
   .form-actions {
     width: 100%;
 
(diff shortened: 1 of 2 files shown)


kevinchappell (approved):
i think youre right about moving the mobile styles to their own file. This is a good fix for now.

Issue #854, opened 2018-11-05
Icon for custom controls not display in mobile version


AndriyYushko:
Icon for custom controls not display in mobile version: http://joxi.ru/Grq7Dq5iQl0wp2

lucasnetau:
A fix for this has been raised in PR https://github.com/kevinchappell/formBuilder/pull/1408

PR #1408, opened 2023-09-11
Fix GH-854 Show icon for custom controls on mobile phone screen sizes

lucasnetau:
When in small screen mode, ensure that the .control-icon is visible and sized the same as .frmb-control li::before

Adjust inputSet control to have the same element layout as custom controls to ensure CSS selectors apply

Fixes: #854


kevinchappell (approved):
awesome! i love these mobile improvements.

Issue #1048, opened 2020-03-11
Drop down is not rendering when JSON data is provided in Form Render:


ammaran2000:
### Description:
Hello, I am rendering the form with JSON data. Dropdown list is not rendering properly and is expended as a list like I can select the multiple values. On my investigation, I found out that this element ""multiple":false" is creating the problem and when I remove this then its working fine. Can you please guide me how to remove this glitch or what is way around other than using xml data instead of JSON data?
Rendering from the XML data is working file.

### Environment Details:
Windows 10 Enterprise
Simple HTML Page

 - formBuilder Version: 3.3.5
 - Browser: Version 76.0.3809.87 (Official Build) (64-bit)
 - OS: Windows 10 Enterprise (64 bit)

### Expected Behavior
Drop down should be rendered as

buttle:
Hi,

I can confirm this (though I haven't tested it with XML like @ammaran2000 says).

Jojoshua:
Try with the latest version. Feel free to reopen if still an issue

Issue #727, opened 2018-03-16
same class name problem ("form-group")


ug-code:
Hi kevinchappell,

Sorry for my english 
my english is very bad : )

your javascript file is not custom for my theme [metronic](https://keenthemes.com/metronic/preview/?demo=default)

for example 
my theme css use  this class name  (form-group)
``<div class="form-group"> bla bla</div> ``

and you javascript file is as follows (form-builder.min.js)
``
.form-elements, .form-group, .multi-row span, textarea {
    display: block; 
}
``

you have to add custom name  
for example 
``
.fb-form-elements, .fb-form-group, .fb-multi-row span, fb.textarea {
    display: block; 
}
``
or
``
.fb.form-elements, .fb.form-group, .fb.multi-row span, fb.textarea {
    display: block; 
}
``

bretto36:
This is still present and an issue. In bootstrap 4 form-group is by default display:flex but this class forces it to be display block. 

Can we simple add a prefix to the class names

`.form-builder .form-elements, .form-builder .form-group, .form-builder .multi-row span, .form-builder textarea { display: block; }`

dainw:
Including formBuilder on a bootstrap page results in all form-group elements being broken with display:block. As an example, the button on this form should be attached to the right side of the search box:

![image](https://user-images.githubusercontent.com/6842359/166292015-61933c07-00d5-4ae0-ae78-76a0db143a55.png)

It looks like [form-builder.scss](https://github.com/kevinchappell/formBuilder/blob/master/src/sass/form-builder.scss) might be closing the .form-wrap.form-builder block early. Please reference @sophanox [comment](https://github.com/kevinchappell/formBuilder/issues/594#issuecomment-395476193) for a possible solution.

lucasnetau:
duplicate of https://github.com/kevinchappell/formBuilder/issues/594

Issue #594, opened 2017-08-30
formBuilder seems to be too liberal in re-styling basic types


jhwinters:
### Description:
Simply including the formbuilder.min.js file in your web page, without even invoking it, is enough to get extra stylesheets injected into your page, re-defining your layout.

The particular problem which I hit is that it re-styles "select" elements to have a margin-bottom of 0px, over-riding a setting which I made earlier in my style definitions.

Surely the selector which it uses should be more selective - it should affect only such elements within formBuilder forms - not everywhere in the application?

### Environment Details:

 - formBuilder Version: 2.5.3-4
 - Browser: Chrome
 - OS: Linux

### Expected Behavior
formBuilder's styles should be sufficiently specific that they affect only its own elements.

### Actual Behavior
Currently, they affect the whole application.

### Steps to Reproduce
As above.

### Screenshot - *(optional)*

![screenshot from 2017-08-30 10-00-54](https://user-images.githubusercontent.com/345516/29864337-40f9c2ae-8d6a-11e7-8cfc-34ca8b2cdcf2.png)

Note how the drop-downs have been squished together.  The Chrome debugger indicates that the change has been imposed by formBuilder's injected CSS, even though this isn't a formBuilder form.  The effect is achieved simply by including formBuilder's js file, without invoking it in any way.

wnoveno:
I've had the same problem as well. 
A suggestion to the maintainers is to create a separate CSS file, it would also have the added benefit of custom styling.

jhwinters:
I agree with the idea of having a separate, explicitly included, CSS file.  It may seem superficially attractive to have the JS code inject its own embedded stylesheet, thus meaning that the package can be distributed as a single file, but it's not standard behaviour and it will inevitably lead to problems like this.

Possible fixes:

1. Make the CSS styles more specific, so they affect only elements within a formBuilder parent element (this should be done regardless).

2. Don't inject the stylesheet unless and until formBuilder code is invoked.  It thus won't do any damage on other pages of an application.  Not really a proper fix, but possibly easier to do quickly.

3. Remove the stylesheet from the JS file entirely, and put it where it should be - in a CSS file.  This then means that end users can tweak it if necessary.

sophanox:
#602 is still causing me issues. In my case it's overriding the bootstrap classes:

![image](https://user-images.githubusercontent.com/19409202/40984025-d48cf1ea-68d8-11e8-8afd-b7a949ed81b6.png)

[...]

PR #1409, opened 2023-09-11
CSS cleanup of unused selectors and small issues

lucasnetau:
@kevinchappell This is the start of some CSS cleanup. 

Once #1407 and #1048 are merged then I have a pending fix for the old issues #727 and #594


src/sass/form-builder.scss
```diff
@@ -124,93 +116,4 @@ textarea {
...
-    margin-top: 10px;
-    text-align: right;
-    clear: both;
-
-    .btn {
-      margin-left: 10px;
-    }
-  }
-}
+@import 'dialog';
```
kevinchappell, line 119: :+1:

kevinchappell (approved):
good call separating out the dialog style

Issue #1353, opened 2023-01-16
Switch to sass from node-sass breaks my icons for some reason.


erpdatallc:
### Description:
On my little web app (expansion of the standard demo) that I developed using v3.6.1 it has been working fine, tried upgrading to 3.8.3 and the icons were replaced with "funky text".  Traced it to breaking at 3.7.4 when the sass processor was changed from "node-sass" to "sass".
### Environment Details:

 - formBuilder Version: 3.7.4+
 - Browser: Chrome
 - OS: Windoze

### Expected Behavior
Icons on the formbuilder
### Actual Behavior
Crazy text instead of icons
### Steps to Reproduce
Under sass processor, the form-builder.min.js entries for icons that are dynamically loaded are of the form:
.formbuilder-icon-autocomplete:before{content:"î €"}
.formbuilder-icon-date:before{content:"î "}

When I rebuild 3.8.3 using node-sass instead I get the old behaviour in that section of the js:
.formbuilder-icon-autocomplete:before{content:'\\\\e800'}
.formbuilder-icon-date:before{content:'\\\\e801'}

lucasnetau:
Do you have <meta charset="utf-8"/> in your HTML document? 

Looking a the webpack output between development and production builds, Webpack CSS Loader appears to be dropping the `@charset "UTF-8";` declaration of the injected CSS styles. It's possible that Windows is picking the wrong encoding, the meta tag should give the appropriate hint for the styles.

lucasnetau:
@erpdatallc I was able to replicate this and adding `<meta charset="utf-8"/>` into my HTML head section resolved the issue. It's a required tag for libraries like Bootstrap.

lucasnetau:
Related upstream issues
https://github.com/webpack-contrib/sass-loader/issues/973
https://github.com/sass/dart-sass/issues/1387

The correct way to handle this is to either include `<meta charset="utf-8"/>` into the HTML document or load form-builder.min.js with a charset assigned `<script src="form-builder.min.js" charset="UTF-8"></script>`

lucasnetau:
Documentation should be updated to ensure that the charset is configured correctly. For example Bootstrap requires this tag to be set https://getbootstrap.com/docs/5.3/getting-started/introduction/#quick-start

PR #1411, opened 2023-09-13
Update docs to fix UTF-8 encoding issues

lucasnetau:
Since the SASS processing was changed from node-sass to sass, the @charset=utf-8 for the styles is not carried through to the production minimised css. On systems where the HTML page is not with UTF-8 encoding set the css icons will be rendered incorrectly. The meta tag 'charset' should be set to utf-8, this is the same requirement as CSS libraries such as Bootstrap.

Update Getting Started documentation with a note about ensuring the charset for the document is set to utf-8 to ensure the CSS icon styles render correctly when the local charset is different

See: https://github.com/webpack-contrib/sass-loader/issues/973
See: https://github.com/sass/dart-sass/issues/1387
Closes: #1353 

Note: Bumped jQuery versions used in demo as any straight copy of the getting started will result in old and outdated versions being used by default.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -14,16 +14,20 @@ Once you have the module/plugin installed, include it in your build process for 
 
 ## Basic Example
 
+**Note**: The document charset should be set to utf-8 to ensure CSS font styles are rendered correctly 
+
 ```html
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
-    <title>Example formBuilder</title>
+  <!-- Required meta tag -->
+  <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
excellent. how did you find this issue?

lucasnetau:
> excellent. how did you find this issue?

The PR reminded me of the encoding issue with windows defaulting to a different encoding than UTF-8. I could see that the charset was correctly shown when running a development build but not a production build. I did a search on dart-saas and saas-loaded for UTF-8 and found those issues above, then remembered that Bootstrap requires the meta tag.

PR #1412, opened 2023-09-13
Ensure we remove bootstrap classes for child elements in the field

lucasnetau:
col- and row- classes were not removed from fields that are more complex than a simple input element (e.g. checkbox-groups).

tmpCleanColumnInfo() could be called multiple times leading classnames to have tmp- appended multiple times (eg tmp-tmp-row-1), ensure we don't keep applying the change by removing any applied tmp- values before reapplying

tmpCleanColumnInfo would append tmp- to any classname found containing the word 'row', adjust replacement to only replace row- values

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -854,7 +854,7 @@ export default class Helpers {
       cleanResults.forEach(result => {
         if (result['columnInfo'].columnSize) {
           const currentClassRow = rowContainer.attr('class')
-          if (currentClassRow != result['columnInfo'].columnSize) {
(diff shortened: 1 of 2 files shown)


src/js/helpers.js
```diff
@@ -1288,13 +1288,16 @@ export default class Helpers {
     })
 
     function tmpCleanColumnInfo($field) {
-      var classAttr = $field.attr('class')
```
kevinchappell, line 1291: oof, that was old

kevinchappell (approved):

Issue #915, opened 2019-02-26
typeUserAttrs does not take effect when the field is first added


baozhoutao:
### Description:
In the Usage of https://formbuilder.online/docs/formBuilder/options/typeUserAttrs/, add the date type field. The first added date type field does not have the min,max attribute.
### Environment Details:

 - formBuilder Version: 3.1.3
 - Browser: chrome 71.xxx
 - OS:  win10

Mordef:
Got similar:
`const typeUserAttrs = {
    text: {
        dsgvo: {
            label: 'DSGVO',
            value: false,
        }
    }
};`
When insert a textfield i got a textfield, not a checkbox.
On load JSON first, with "dgsvo":true its displayed with checkbox

rcbeton:
I have the same issu.

rcbeton:
I fund the solution for this. You need to set the value attr to get i to get it show on first add.

AndriyYushko:
I also have this problem

MarcusDelvecchio:
value = "" did it for me, wish I saw this before I had the problem though.

lucasnetau:
There are two issues here:

@baozhoutao 

Per the doco 

> Attribute input types are generated using the typeof for the value.

So if there is no 'value' set for the userAttr then userAttrType() returns 'undefined' and the field isn't added.

One to issue here:
* Demo needs to be fixed because this demo doesn't work
* userAttrType() suggests that it should return 'string' as the type if value is undefined per `find(typeCondition => typeCondition[1](attrData))[0] || 'string'` however 'undefined' is a string from typeof so it's never false or null

@Mordef 
> Got similar: `const typeUserAttrs = { text: { dsgvo: { label: 'DSGVO', value: false, } } };` When insert a textfield i got a textfield, not a checkbox. On load JSON first, with "dgsvo":true its displayed with checkbox

This is due to a conversion of the value from _false_ to an empty string.

I can likely fix this with the following change
```javascript
const origValue = tUA.value || (attrValType === 'boolean' ? false : '')
tUA.value = values[attribute] || tUA.value || (attrValType === 'boolean' ? false : '')
```

However the documented way is to assign the value type: 'checkbox' to your config `const typeUserAttrs = { text: { dsgvo: { label: 'DSGVO', value: false, type: 'checkbox',} } };

lucasnetau:
PR #1413 raised for @Mordef's issue

lucasnetau:
[...]

PR #1413, opened 2023-09-14
Ensure we do not corrupt the userTypeAttr definition when encountering a boolean false value

lucasnetau:
When a user configured a userTypeAttr with `value: false` but no `type: checkbox` the origValue calculation in processTypeUserAttrs would set the  _false_ to _''_ , this would cause the typeUserAttrs value definition to be changed when resetting the tUA.value to the original value.

This had the affect that when additional fields are added with this typeUserAttrs, processTypeUserAttrs would see the value as type string and output a textbox instead of a checkbox.

value: true would be detected as boolean and always kept as boolean. This PR ensures value: false works the same way.

One of the issues raised in https://github.com/kevinchappell/formBuilder/issues/915

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -697,8 +697,8 @@ function FormBuilder(opts, element, $) {
         if (attrValType !== 'undefined') {
           const orig = mi18n.get(attribute)
           const tUA = typeUserAttr[attribute]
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
good fix. thanks for explaining in the description

PR #1414, opened 2023-09-14
Warn when a typeUserAttr definition is unable to be processed instead of silently ignoring

lucasnetau:
Issue raised in https://github.com/kevinchappell/formBuilder/issues/915

The codepen on https://formbuilder.online/docs/formBuilder/options/typeUserAttrs/#example-number-inputs defines min and max fields without setting a value: for processTypeUserAttrs to detect the type. The definition is then silently ignored instead of raising a warning

Additionally the fallback check in userAttrType can not be called because the last check `() => true` will always be true and `typeof attrData.value` will always return a string value, additionally the subType user attribute function expects this to return the 'undefined' string from typeof to trigger subType processing so a fallback is not expected


kevinchappell (approved):
:+1:

Issue #1324, opened 2022-08-17
Question: How do I prevent HTML in labels?


robv8r:
How do I prevent users from adding HTML to a Form Field Label?

There's a configuration option called [disableHTMLLabels](https://formbuilder.online/docs/formBuilder/options/disableHTMLLabels/).  The description of the configuration option states that it "Disables HTML labels."  Setting this value to true or false has no effect on adding HTML to labels.

The CodePen on the [disableHTMLLabels](https://formbuilder.online/docs/formBuilder/options/disableHTMLLabels/) page allows you to create hyperlinks in a field's label.

To recreate the issue:

1. Visit the disableHTMLLabels documentation page.
2. In the CodePen, drag a Form Field onto the Form Builder.
3. Edit the label by clicking the Pencil icon.
4. Change the value to `<a href="https://www.microsoft.com" target="blank">Autocomplete</a>`
5. Click "Close"
6. Click the Label that was just modified
7. A new tab opens showing Microsoft's webpage.

The preferred behavior would be to either disallow HTML in the label or escape the HTML so it isn't rendered as HTML.

If I'm missing something, I'd be grateful if someone could educate me.

robv8r:
Thank you for creating formBuilder and making it a successful open source project.

I fully understand (and respect) that projects like formBuilder are often created and maintained by volunteers.

With respect, I ask if anyone has input on my question above.

Thank you.

lucasnetau:
@kevinchappell , what are your thoughts on this issue?

lucasnetau:
@robv8r PR #1416 has been raised which should complete the implementation of text only labels. This only forces HTML to be treated as text. If you wish to strip HTML I would suggest implementing DOMPurify in your code.

PR #1416, opened 2023-09-15
Complete disableHTMLLabels feature

lucasnetau:
disableHTMLLabels in formBuilder only changed .fld-label from a contenteditable to input[type=text]. HTML could still be written into the input and the keyup/change event handler would update the label as HTML. formBuilder would also render any html in labels set in formData on initialisation of the form.

Additionally formRender had no support for disableHTMLLabels and therefore would render the labels as HTML when rendering userData

This PR completes the intent of the disableHTMLLabels feature

- Add disableHTMLLabels option to formRender
- layout class now accepts a third optional constructor parameter disableHTMLLabels. Extended classes of layout should take this into account
- Ensure disableHTMLLabels option is honoured in the three places where the label is rendered (layout.label(), formBuilder.appendNewField(), d.stage keyup/change event handler for label .fld-label). This is done by adding the label text into a Text node when constructing the label element and assigning to textContent in the event handler 

Fixes: https://github.com/kevinchappell/formBuilder/issues/1324

### DIFF
--- a/docs/formRender/options.md
+++ b/docs/formRender/options.md
@@ -5,6 +5,7 @@ var defaults = {
   container: false,
   formData: false,
   dataType: 'json', // 'xml' | 'json'
+  disableHTMLLabels: false,
   label: {
     formRendered: 'Form Rendered',
     noFormData: 'No form data.',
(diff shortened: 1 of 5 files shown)


lucasnetau:
@kevinchappell I've marked this as a minor feature version bump because it adds a new config to formRender and changes the class signature of layout (in a backwards compatible way), please let me know if that is ok or I should mark as a fix?

kevinchappell (approved):
You are on a roll. Good work!

Issue #665, opened 2017-11-18
Disable drag and drop for selected elements


rssluca:
Hi all,

I would like to lock the position of the first two elements (loaded with prepend and defaultFields).

Is there a way to do this?

Thanks,

Luca

PR #1417, opened 2023-09-19
Fix insertion position for new controls when append or prepend options are enabled

lucasnetau:
This PR fixes the following issues:

- Dragging a control onto the stage before a prepended field or after a appended field would show the drop placeholder and then silently cancel the drop on cursor release. We can fix this by relocating the placeholder with the jquery-ui.sortable change event
- Adding a field by clicking on the control would insert the new control after a prepended field. Correctly set h.stopIndex to ensure controls are added before the prepended field
- When sorting stage controls using the mobile sort buttons the sort handler did not take into account appended/prepended fields

Partial fix for #665

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -2326,12 +2326,12 @@ function FormBuilder(opts, element, $) {
     let swap
     if ($(evt.target).hasClass('sort-button-higher')) {
       swap = currentItem.prev('li')
-      if (swap.length) {
+      if (swap.length && !swap.hasClass('form-prepend')) {
         currentItem.insertBefore(swap)
       }
     } else {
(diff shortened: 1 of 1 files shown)


src/js/form-builder.js
```diff
@@ -131,6 +131,13 @@ function FormBuilder(opts, element, $) {
       revert: 150,
       beforeStop: (evt, ui) => h.beforeStop.call(h, evt, ui),
       distance: 3,
+      change: function(event, ui) {
+        if (opts.prepend && ui.placeholder.index() < 1) {
```
kevinchappell, line 135: smart

kevinchappell (approved):
much needed polish, good job

PR #1419, opened 2023-09-19
Implement built-in XSS and DOM Clobbering protection

lucasnetau:
Sanitizer is implemented with support for different backends: the experimental Sanitizer API (currently supported by Chrome and Edge), DomPurify, and a fallback jQuery sanitizer implementation, each will be tried in order until one is successful. DomPurify need to be included prior to formBuilder/formRender being initialised.

Sanitizer and DomPurify can be configured by passing in a new configured instance of their respective Class via the sanitizerOptions option. The fallback function can be overridden with a new function the same way.

Backends can be disabled by passing in a false value for that backend via the sanitizerOptions option. All sanitization can be disabled by setting all backends to false

sanitizerOptions = {
  sanitizer: false,
  dompurify: false,
  fallback: false
}

Each backend has been tested against a variety of inputs from https://github.com/mjfroman/moz-libwebrtc-third-party/blob/980ec542e93f88658089ee4ac78fc3dcfd066847/blink/web_tests/wpt_internal/sanitizer-api/support/testcases.sub.js and DomPurify

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -35,7 +35,7 @@ import {
   getContentType,
   generateSelectorClassNames,
 } from './utils'
-import { setElementContent, setSanitizerConfig } from './sanitizer'
+import { attributeWillClobber, setElementContent, setSanitizerConfig } from './sanitizer'
 import fontConfig from '../fonts/config.json'
 const css_prefix_text = fontConfig.css_prefix_text
 
@@ -1714,10 +1714,9 @@ function FormBuilder(opts, element, $) {
     $valWrap.toggle(e.target.value !== 'quill')
   })
 
-  const testForm = document.createElement('form')
   $stage.on('change', '[name="name"]', e => {
     const name = e.target.value
-    if (name in document || name in testForm) {
+    if (attributeWillClobber(name)) {
       //@TODO Notify the user of this potential issue
       opts.notify.error('Potential for Dom Clobbering with field name ' + name)
     }
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -35,6 +35,7 @@ import {
   getContentType,
   generateSelectorClassNames,
 } from './utils'
+import { setElementContent, setSanitizerConfig } from './sanitizer'
 import fontConfig from '../fonts/config.json'
 const css_prefix_text = fontConfig.css_prefix_text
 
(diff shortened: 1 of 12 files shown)


lucasnetau:
Hi @kevinchappell , this is a WIP and not ready for merging yet. There is an outstanding issue around notifying the user of possible Dom Clobbering when the 'name' field is set in FormBuilder, and additionally removing Dom Clobbering attributes in FormRender that could affect a `<form>` element container the rendered form

src/js/form-builder.js
```diff
@@ -65,6 +66,13 @@ function FormBuilder(opts, element, $) {
   formBuilder.colWrapperClass = colWrapperClass
   formBuilder.fieldSelector = opts.enableEnhancedBootstrapGrid ? rowWrapperClassSelector : defaultFieldSelector
 
+  //Initialise HTML sanitizer
+  setSanitizerConfig(opts.sanitizerOptions)
+  if ($(element).closest('form').length) {
+    //Due to Dom Clobbering potential with the stage and the lack of requirement for a Form element, warn for this type of setup
```
kevinchappell, line 72: what is the potential outcome if formBuilder is loaded in a `form` tag?

src/js/form-builder.js
```diff
@@ -93,11 +101,7 @@ function FormBuilder(opts, element, $) {
   let cloneControls
 
   function enhancedBootstrapEnabled() {
-    if (!opts.enableEnhancedBootstrapGrid) {
-      return false
-    }
-
-    return true
+    return !!opts.enableEnhancedBootstrapGrid
```
kevinchappell, line 104: :+1:

src/js/form-builder.js
```diff
@@ -1702,6 +1706,15 @@ function FormBuilder(opts, element, $) {
     $valWrap.toggle(e.target.value !== 'quill')
   })
 
+  const testForm = document.createElement('form')
+  $stage.on('change', '[name="name"]', e => {
+    const name = e.target.value
+    if (name in document || name in testForm) {
+      //@TODO Notify the user of this potential issue
```
kevinchappell, line 1713: i dont quite understand the issue here. wouldn't we _want_ to dom clobber names out of the html?

kevinchappell (commented):
an excellent feature and the warnings are helpful. im looking forward to this one

src/js/form-builder.js
lucasnetau: FormBuilder has inputs in the edit panel with name=className and name=name, these clobber form.className form.name.

src/js/form-builder.js
lucasnetau: From https://whatpr.org/whatwg_html/8474/form-control-infrastructure.html#attributes-common-to-form-controls

> DOM clobbering is a common cause of security issues. Avoid using the names of built-in form properties with the [name](https://whatpr.org/whatwg_html/8474/form-control-infrastructure.html#attr-fe-name) and id content attributes.

Say if you name your Text input 'method' when we render this field in formRender (not preview since fields name have a suffix -preview appended) then the containing form's method attribute will be clobbered with the text input

kevinchappell: i see. thanks for explaining

src/js/form-builder.js
kevinchappell: sorry i missed that explanation in chat at first, thanks for repeating here.

lucasnetau:
@kevinchappell , this is good to go. As we discussed protections are disabled by default in this release. Documentation has been added

kevinchappell (approved):
awesome work. merging!

PR #1421, opened 2023-09-21
Fix regression from #1412

### LINKED ISSUE #1420: Adding Autocomplete results to "t.querySelectorAll is not a function"
### Description:
When i simply add a new Autocomplete with Form Builder an error appears.

### Environment Details:

 - formBuilder Version: 3.10.0
 - Browser: Mozilla
 - OS: Windows 10

### Actual Behavior
t.querySelectorAll is not a function

### Steps to Reproduce
- In Form Builder click on "Autocomplete" to add a new autocomplete

### Screenshot - *(optional)*

image

lucasnetau:
Autocomplete builds the field config as an Array of HTMLElements, handle this case when processing the className

Fixes #1420

### DIFF
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -19,11 +19,15 @@ const processClassName = (data, field) => {
       field.classList.remove(...classes)
     }
 
-    field.querySelectorAll('[class*=row-],[class*=col-]').forEach(element => {
+    //field may be a single element, dom fragment, or an array of elements
+    if (!Array.isArray(field)) {
+      field = [field]
+    }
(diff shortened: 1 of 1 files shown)


lucasnetau:
This was not picked up in automated tested due to the addField action takes a different code path to clicking on the control in the panel. When adding via the control panel className is set to 'form-control' 

```javascript
    if (isNew && ['text', 'number', 'file', 'date', 'select', 'textarea', 'autocomplete'].includes(field.type)) {
      field.className = field.className || 'form-control'
    }
```

however when adding via addField this is not performed. Without a className being set the code issue fixed by the PR is not triggered.

Have added test cases adding fields via simulated Click events

kevinchappell (approved):

PR #1422, opened 2023-09-21
Update deploy flow

kevinchappell:
This PR adds a new workflow that is triggered when the plugin is published. The new workflow checksout the website repo, updates dependencies and the version then pushes the changes and tags which then triggers a release of the website in the kevinchappell/formBuilder-site repo.

With this change, the website will be updated with every release of formBuilder


.github/workflows/deploy-site.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      issues: write
+      pull-requests: write
+      id-token: write
+      packages: write
```
lucasnetau, line 20: I think the only permission you need is this one, the rest can be none

```yaml
    permissions:
      contents: write
```

.github/workflows/deploy-site.yaml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Checkout formBuilder
+        uses: actions/checkout@v3
```
lucasnetau, line 23: actions/checkout@v4 is released

kevinchappell: oh nice, i was not aware. will use v4

PR #1423, opened 2023-09-21
Change Deploy website trigger

lucasnetau:
Due to recursion protection GitHub actions does not trigger the publish event when the release is created via another action. Instead use the workflow_run event to chain Publish with Deploy Website


kevinchappell (approved):
Nice! Good catch about removing current and chaining rocks. Does this change mean deploy-site is run regardless of if a release was cut? Sometimes the publish action will not make a new release unless the commits are formatted to trigger one.

PR #1425, opened 2023-09-25
Improve the definition of utils.trimObj() to not modify the source object

lucasnetau:
All usages of trimObject use the return value however due to working on the source object we modify the original object too. We can tidy up trimObj definition by removing attributes from a copy of the source object

This was found during the automated testing development.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -13,11 +13,15 @@ window.fbEditors = {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
good fix. i reused this util elsewhere and needed to make a similar fix but forgot to backport it here. thanks

PR #1426, opened 2023-09-25
Various small cleanups to code

lucasnetau:
@kevinchappell There are a bunch of little cleanups in this PR that I've been keeping in the testing framework branch but should be applied now rather than with that PR. There are no functionality changes with this.

Main items: 
- Remove calling clearAllFields with an animate flag, flag was removed in https://github.com/lucasnetau/formBuilder/commit/753280f183663dd153802458209b6427c10f4c36
- xmlSave does not support minify, don't call xmlSave with a minify flag set
- Various function parameters not marked as optional
- JSDoc cleanup, the majority of these JSDoc cleanup is to ensure we have the correct definition of the function and ensure that when we are testing it we can test against that contract.

### DIFF
--- a/src/js/control.js
+++ b/src/js/control.js
@@ -119,7 +119,7 @@ export default class control {
 
   /**
    * Looks up the classRegister & returns registered types or subtypes
-   * @param  {String} type optional type of control we want to look up
+   * @param  {string|false} type optional type of control we want to look up
    * subtypes of. If not specified will return all types
    * @return {Array} registered types (or subtypes)
    */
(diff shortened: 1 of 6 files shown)


src/js/control.js
```diff
@@ -134,7 +134,7 @@ export default class control {
       if (type) {
         return key.indexOf(type + '.') > -1
       }
-      return key.indexOf('.') == -1
+      return key.indexOf('.') === -1
```
kevinchappell, line 137: another version, but less readable is `!~key.indexOf('.')`

kevinchappell (approved):
:+1:

PR #1431, opened 2023-10-04
Enhanced bootstrap feature fixes

lucasnetau:
While testing I encountered a variety of small issues with the enhancedBootstrap feature.

* Due to the bootstrap feature placing multiple DIVs within the Stage, the stopIndex calculation was incorrect leading to new elements being placed after any prepend field.
* Drag and dropping Header fields resulted in invalid fields that could not be interacted with until the preview was refreshed due to non-removed row- and col- values
* Drop placeholders were not created when cloning fields
* Rapid movement could lead to a variety of issues with placeholders, either showing when they shouldn't or remaining hidden when they should be shown
* Resizing columns was unusable with a touchpad, refactored the event handler to be less erratic 
* Mobile sort buttons are incompatible with enhancedBootstrap feature so have been disabled when this mode is active.
* Rows moved by keyboard would incorrectly sort above and below the prepend and append fields
* Improved highlight when sorting via mobile buttons (using helpers.toggleHighlight built-in)

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1418,6 +1418,7 @@ function FormBuilder(opts, element, $) {
         if (listFieldItem.length) {
           CheckTinyMCETransition(listFieldItem)
           UpdatePreviewAndSave(listFieldItem)
+          h.tmpCleanPrevHolder($(ui.item).find('.prev-holder'))
         }
       },
       start: function () {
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
awesome fixes. thanks for the detailed explanation.

PR #1432, opened 2023-10-05
Escape formData in showData dialog for all dataTypes

lucasnetau:
formData may contain HTML (eg Labels), need to escape for all dataTypes before adding into the code element, not just for XML datatype

JSON is invalid if copy-paste from showData dialog:

```javascript
       {
           "type": "text",
           "required": false,
           "label": "Item<br>Label",
           "className": "form-control",
       },
```

would be rendered as

```javascript
       {
           "type": "text",
           "required": false,
           "label": "Item
Label",
           "className": "form-control",
       },
```

### DIFF
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -988,11 +988,7 @@ export default class Helpers {
    * Open a dialog with the form's data
    */
   showData() {
-    let formData = this.getFormData(config.opts.dataType, true)
-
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
good fix. do you think its worth supporting the xml datatype still? does anyone use it?

PR #1433, opened 2023-10-05
Refactor CSS to fix various namespace clashes and improve compatibility with newer Bootstrap CSS libraries

### LINKED ISSUE #594: formBuilder seems to be too liberal in re-styling basic types

### LINKED ISSUE #727: same class name problem ("form-group")

lucasnetau:
FormBuilder has an embedded Bootstrap 3 stylesheet which can cause clashes with HTML elements outside of the formBuilder and formRender containers.

This PR does some refactoring of the style and introduces a modification to an option flag to selectively disable Bootstrap 3 styles into the form.

Refactoring: Move classes under the .rendered-form and .form-wrap.form-builder containers. This ensures we are only setting styles for Elements within our container.

Fixes #594
Fixes #727 

Refactoring: Tooltips styles de-duplicated and moved under their own includable scss file to share between form-render and form-builder

Feat: By moving the embedding of Bootstrap 3 styles under an enabling class .formbuilder-embedded-bootstrap we can selectively disable those styles by removing that class. This can be done by setting `disableInjectedStyle` to `'bootstrap'`, effectively changing `disableInjectedStyle` from a boolean to a boolean|string option. As disableInjectedStyle defaults to false, the Bootstrap classes will be embedded

Testing has shown no oddities on either Builder nor Render.

### DIFF
--- a/docs/formBuilder/options/disableInjectedStyle.md
+++ b/docs/formBuilder/options/disableInjectedStyle.md
@@ -1,7 +1,19 @@
 # disableInjectedStyle
-I'm not sure why anyone would want to do this but it's been requested a couple times so here it is... `disableInjectedStyle` enables you to disabled the injeted style that ships with the plugin.
 
-## Usage
+FormBuilder ships with a minimal Bootstrap 3 set of CSS classes for rendering form controls. You may choose to disable these to ensure they do not clash with a Bootstrap 4 or 5 included stylesheet
+
+## Disable Embedded Bootstrap Classes
+```javascript
+var options = {
+  disableInjectedStyle: "bootstrap"
+};
+$(container).formBuilder(options);
+```
+
+## Disable all embedded styles
+
+You can also completely disable all embedded styles, I'm not sure why anyone would want to do this, but it's been requested a couple of times so here it is... `disableInjectedStyle` enables you to disable the injected style that ships with the plugin.
+
 ```javascript
 var options = {
   disableInjectedStyle: true
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -2414,9 +2414,11 @@ function FormBuilder(opts, element, $) {
 
   loadFields()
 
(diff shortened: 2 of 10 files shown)


kevinchappell (approved):
changes look good to me

PR #1435, opened 2023-10-05
fix: fallbackSantizer used before defined

kevinchappell:
fixed by converting to declared function instead of expression to leverage hoisting

### DIFF
--- a/src/js/sanitizer.js
+++ b/src/js/sanitizer.js
@@ -8,7 +8,7 @@ const sanitizerConfig = {
     form: true,
     namespaceAttributes: false, //whether to prefix with user-content-
   },
-  backendOrder: ['dompurify','sanitizer','fallback'],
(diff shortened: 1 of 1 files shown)


src/js/sanitizer.js
```diff
@@ -64,15 +64,15 @@ export const isPotentiallyDangerousAttribute = (attrName, attrValue) => {
...
+  attrValue = attrValue ? attrValue + '' : ''
   return (
     attrNameLc.startsWith('on')
     || ['form', 'formaction'].includes(attrNameLc)
     || attrValue.trim().toLowerCase().startsWith('javascript:')
   )
 }
 
-const fallbackSanitizer = content => {
+function fallbackSanitizer(content) {
```
kevinchappell, line 75: this is the fix

PR #1436, opened 2023-10-05
fix: missing labels

kevinchappell:


lucasnetau:
Opened #1437 to include this fix plus another one with setup of Sanitizer

YANGSANGHEE:
Please apply it to CDN, too

PR #1437, opened 2023-10-05
fix: Setup of fallBack Sanitizer and processing of content when sanitizer is not enabled

lucasnetau:


kevinchappell (approved):
thanks for the quick fix!

PR #1438, opened 2023-10-06
Introduce Jest automated testing into formBuilder

lucasnetau:
Initial commit to bring in dependencies and setup folder structure and GitHub action


tests/setup-jest.js
```diff
@@ -0,0 +1,37 @@
...
+  open            : jest.fn(), //(method, url) => {console.log(method + ':' + url)},
+  send            : jest.fn().mockImplementation(function(){ this.onload() }),
+  setRequestHeader: jest.fn(),
+  abort: jest.fn(),
+  addEventListener: jest.fn(),
+  dispatchEvent: jest.fn(),
+  error: jest.fn(),
+  getAllResponseHeaders: jest.fn(),
+  getResponseHeader: jest.fn(),
+  //load: jest.fn(),
```
kevinchappell, line 18: do we need this?

lucasnetau: So far yes.

The jQuery part sets up jQuery into the global space

Since I am testing the raw .js files rather than the built files I need to configure FB_EN_US to set the language (what webpack normally does)

I needed the xhrMockClass when testing loading of script etc. We can likely adjust in the future but mocking these helps

Issue #962, opened 2019-06-07
Unable to save data of rendered form- Rating control plugin 


syed-asim:
### Description:
Thanks Kevin, for creating this wonderful tool.

I am using the the rating plugin with typeUserAttrs (custom attribute) "ratingData". 
`typeUserAttrs: {
        starRating: {
            ratingData: {
                label: 'Rating',
                value: '5'
            } } }`

However, once i render the form > fill in the form (change the rating). And then save the userData using:

$('#' + SurveyPages[currentPageIndex].PageId).formRender('userData'); 

The above return the data perfectly, but the value of typeUserAttrs (custom attribute) "ratingData"  remains the same i.e. 5. Below is the onRender method for rating plugin.

 ```
onRender() {
            var el = this;

            $('#' + el.config.name).rateYo({

                onSet: function (rating, rateYoInstance) {
                    $(rateYoInstance).next().val(rating);

// tried setting the config
                    el.config.ratingData = rating;
// tried setting the  raw config
                    el.rawConfig.ratingData = rating;
// tried setting the value as well
                    el.config.value = rating;
                },
                rating: el.config.ratingData || 3,
                numStars: el.config.noStars || 5,
                fullStar: true,
            });

        }
``` 

Please advise how to get the updating value. Any help would be really appreciated.

I searched google and  checked all the issues as well, but couldn't find a fix for it that is why i am reporting here.

### Environment Details:

 - formBuilder Version: 
 - Browser: 
 - OS: 

### Expected Behavior
$('#' + SurveyPages[currentPageIndex].PageId).formRender('userData'); 

### Actual Behavior
Return the default value

### Screenshot - *(optional)*
![image](https://user-images.githubusercontent.com/36839379/59111756-ba740500-8952-11e9-9c2d-e029c57b667d.png)

iftkharhussain:
we're on the same page, please update if you found any solution. Thanks

PR #1446, opened 2023-10-09
Complete implementation of starRating

lucasnetau:
starRating plugin did not have the logic to load userData into the control, nor did it save the rating anywhere for getUserData to retrieve. This PR completes the implementation of starRating

Fixes: #962

@kevinchappell Was starRating meant to be implemented in this way as an example or was this an oversight?


kevinchappell (approved):
👍️

Issue #611, opened 2017-09-13
how to add conmmon attribute into each field?


baoyadong:
As the title said,I want to add common attribute into each field.In my program ,I want to add a 
select label into all the fields,how can it easily add?
   for example ,I want to add 'groupSort' a select label,how can I add  in all the fields?
code ie:
groupSort:{
        label:'组排序',
        type:'select',
        options:[
        {'label': '1'},
        {'label': '2'},
        {'label': '3'},
        {'label': '4'}
          ]
        }
    },

bstaniszczak:
I need this feature to.

baoyadong:
I also find that I cannot add the attribute of select label to the field very well ,it cannot show the content of option ,the browse shows code like this : 
<select id="groupSort-frmb-1505288748534-fld-1" title="组别" name="groupSort" class="fld-groupSort form-control">
    <option value="0" selected="null"></option>         
    <option value="1"></option>
    <option value="2"></option>
    <option value="3"></option>
</select>
I don't know why.Can you help me solve this problem as soon as possible?

bstaniszczak:
I do it like this:
Define array with types:

var available_types = [{type:'text'},{type:'button'},{type:'select'},{type:'button'},{type:'checkbox-group'},{type:'date'},{type:'file'},{type:'header'},{type:'hidden'},{type:'paragraph'},{type:'number'},{type:'radio-group'},{type:'select'},{type:'text'},{type:'starRating'},{type:'autocomplete'},{type:'textarea'}];

and in loop build user type attrs:

  var typeUserAttrs = {};
  for(e in available_types){
	  typeUserAttrs[available_types[e].type]={parentClass: {label: 'Kolumny',options: {'col-md-3': '1/4','col-md-6': '1/2','col-md-9': '3/4','col-md-12': '1',}}};
  }

not nice but work for me.

baoyadong:
OK,I tried ,and it worked .Thank you very much! @bstaniszczak

bstaniszczak:
no problem, it would be nice to get list of all types from app.

baoyadong:
Yes ,I think it's a good idea!

kevinchappell:
Will add a `userAttrs` option which applies the attributes to all fields and an action/method for registered controls

rachelSwimmer:
Hi,
if I want to create a typeuserattrs with my html,
can I do something like that?

sivashree:
[...]

PR #1447, opened 2023-10-11
feat: Support typeUserAttrs and typeUserEvents for all types with a wildcard config key

lucasnetau:
Allow you to set a typeUserAttrs on all types and typeUserEvents on all types using the wildcard asterisk *. 

config applied too a specific type will override config applied to the wildcard (ie onadd for key text will override onadd for wildcard, attribute by name will override an attribute by the same name in wildcard config)

Tests and documentation updates included

@kevinchappell I choose using a wildcard setting rather than creating a userAttrs and userEvents config key, however please advise if you prefer creating new config options 

Fixes #https://github.com/kevinchappell/formBuilder/issues/611

### DIFF
--- a/docs/formBuilder/options/typeUserAttrs.md
+++ b/docs/formBuilder/options/typeUserAttrs.md
@@ -64,5 +64,17 @@ const typeUserAttrs = {
 };
 ```
 
+## Example Input for all types
+```javascript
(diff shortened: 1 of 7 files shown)


kevinchappell (approved):
awesome

PR #1449, opened 2023-10-13
Unify the regular and enhancedBootstrap UX

lucasnetau:
This consolidates code between the regular and enhancedBootstrap formBuilder interfaces. It removes a lot of duplicate code and different styles.

Visually, regular formBuilder now has sortable placeholders that look the same as the ones in bootstrap, including when sorting fields on the stage. This improves UX as you can clearly see where elements will be dropped

In enhancedBootstrap, a lot of work was done into reducing the visible jumping that occurred when you dragged elements around, drop elements we also improved for existing rows with one column, previously you could have drop zones appear above and below the first column in a row, now they should appear on and left and right as expected.

Tests have been added for bootstrap functions in Helpers.js

The stage is changed to a display: flex container so that we can always have a visible drop zone taking up the remaining space on the stage. We use flex within the stage to support bootstrap so this should have minimal affect

summary
* Consolidate controls.sortable configuration into a single definition
 * Prevent control jumping by control + helper buttons invisible instead of removing them
 * Make stage a flex container and ensure the last drop zone is a flex-grow so that it spans remaining space in the stage
 * Use boostrap hover class in regular formBuilder to give additional user feedback on drop zones for sorting
 * Various equality checks from == to ===
 * stopIndex fixes for bootstrap function

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -111,7 +111,6 @@ export const config = {}
 export const gridClassNames = {
   rowWrapperClass: 'rowWrapper',
   colWrapperClass: 'colWrapper',
-  tmpColWrapperClass: 'tempColWrapper',
   tmpRowPlaceholderClass: 'tempRowWrapper',
   invisibleRowPlaceholderClass: 'invisibleRowPlaceholder',
 }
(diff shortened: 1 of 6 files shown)


kevinchappell (approved):
finally got to test this. great work! nice and snappy even with bootstrap

PR #1454, opened 2023-10-18
Overhaul custom control setup

### LINKED ISSUE #1090: Can't add custom element in multiple form builder!
### Description:
I'm using form builder on Laravel blade and created a custom element below.
```javascript
{
   className: "form-control",
   label: "Relationship",
   type: "select",
   values: @json($relationships),
   icon: '<i class="fas fa-link"></i>'
}
```

That custom element is used by multiple form builder within the same page. I'm using multi-page builder as described in [here](https://formbuilder.online/docs/formBuilder/demos/tabs/).

Screenshot from 2020-06-23 11-30-57

### Environment Details:

 - formBuilder Version: 3.4.2
 - Browser: Chrome 83.0.4103.106
 - OS: Ubuntu 20.04

### Expected Behavior
Custom element should be added in every form builder instance.

### Actual Behavior
It's only work in last initiated form builder (in my screenshot above, it's Page 7). When I add in other pages, it show below error.

Screenshot from 2020-06-23 11-26-50

That custom element is imported as `type + random number`. So, it can't find in template and showing above error. Is this the default behavior ? How can I fix that ?

lucasnetau:
Custom controls we loaded globally via static methods and properties in custom.js, each formBuilder instance would remove the old definitions before adding their own in. This meant that field definitions would appear in the controls in earlier formBuilders however would throw an error when clicked because the definition has been removed

This PR reworks the custom controls by introducing a per instance container to register and retrieve custom control registrations. A proxy class is used to map template to the controlCustom class.

Closes #1090

### DIFF
--- a/tests/control/custom.test.js
+++ b/tests/control/custom.test.js
@@ -1,8 +1,9 @@
 require('../setup-fb')
 require('./../../src/js/form-builder.js')
+require('./../../src/js/form-render.js')
 
 describe('Test Custom Control', () => {
-  test('test building custom control element', async () => {
+  test('test add custom field with template', async () => {
     const fbWrap = $('<div>')
     const cbOnRender = jest.fn()
 
@@ -16,7 +17,7 @@ describe('Test Custom Control', () => {
     const templates = {
       starRating: function(fieldData) {
         return {
-          field: '<span id="'+fieldData.name+'">',
+          field: this.markup('span', null, { name: fieldData.name}),
           onRender: cbOnRender
         }
       }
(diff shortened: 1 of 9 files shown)


tests/control/custom.test.js
```diff
@@ -25,10 +26,202 @@ describe('Test Custom Control', () => {
...
+  test('override built-in with template', async () => {
+    const fbWrap = $('<div>')
+    const cbOnRender = jest.fn()
+
+    const fields = [
+      {
+        type: 'checkbox-group',
+        subtype: 'custom-group',
+        label: 'Custom Checkbox Group w/Sub Type',
+        required: !0,
```
kevinchappell, line 78: just curious, why not just `true`?

kevinchappell (approved):
awesome stuff. thanks for fixing this ancient bug

tests/control/custom.test.js
lucasnetau: Can't answer that. I'm not sure if it was my IDE that did that or if I copy/pasted the example from somewhere. I know webpack has a habit of using !0 for true

PR #1456, opened 2023-10-19
Initialise formRender in cases of undefined or empty formData

### LINKED ISSUE #1372: formRender options - setData does not work
### Screenshot
image

lucasnetau:
Ensure we initialise formRender even if there is no field data to render. This enables setData/render to be called after initialisation on an empty form

Fixes #1372

Update documentation for formRender actions

Small fix and tests applied to layout.js for lifting row/col fields

### DIFF
--- a/docs/formRender/actions/render.md
+++ b/docs/formRender/actions/render.md
@@ -1,6 +1,6 @@
 # `render` action
 
-Programmtically render or re-render a formRender instance with new data
+Programmatically render or re-render a formRender instance with new data
 
 Arguments
 | Arg  | Type | Value(s) | Default |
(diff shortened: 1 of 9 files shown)


src/js/form-render.js
```diff
@@ -28,7 +28,7 @@ class FormRender {
       container: false, // string selector or Node element
       dataType: 'json',
       disableHTMLLabels: false,
-      formData: false,
+      formData: [],
```
kevinchappell, line 31: :+1:

docs/formRender/actions/render.md
```diff
@@ -1,6 +1,6 @@
 # `render` action
 
-Programmtically render or re-render a formRender instance with new data
+Programmatically render or re-render a formRender instance with new data
```
kevinchappell, line 3: whoops

kevinchappell (approved):

kevinchappell:
@lucasnetau awesome work! thank you

PR #1458, opened 2023-10-19
Add a helper function to the uninitialised formBuilder instance

lucasnetau:
This change would hopefully stop issues being logged around 'formBuilder.actions.function is not a function TypeError' as it would give feedback that the formBuilder instance is still initialising and they need to wait.

@kevinchappell we could put a URL into the message that takes the developer to the docs around async init? What do you think?

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -2488,6 +2488,8 @@ function FormBuilder(opts, element, $) {
   return formBuilder
 }
 
+const notInitialised = () => { console.error('formBuilder is still initialising') }
+
 const methods = {
   init: (options, elems) => {
     const { i18n, ...opts } = jQuery.extend({}, defaultOptions, options, true)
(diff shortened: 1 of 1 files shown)


kevinchappell:
> This change would hopefully stop issues being logged around 'formBuilder.actions.function is not a function TypeError' as it would give feedback that the formBuilder instance is still initialising and they need to wait.
> 
> @kevinchappell we could put a URL into the message that takes the developer to the docs around async init? What do you think?

That is a great idea. Should we make that message translatable?

lucasnetau:
@kevinchappell, good idea

lucasnetau:
@kevinchappell mi18n may not be initialised by that time, so I'm guessing I need a fallback

kevinchappell:
> @kevinchappell mi18n may not be initialised by that time, so I'm guessing I need a fallback

Ah right of course. English should be fine for this type message because i think mostly developers would see it

kevinchappell (approved):

lucasnetau:
Closed in favour of #1459

Issue #1087, opened 2020-06-14
Multiple forms on the same page


florent-dehanne:
### Description:

Hi,

I try to have multiples forms on the same page (one for each language).
However, when I try to get data, it's always the data from the last form.

Thanks.

### Environment Details:

 - formBuilder Version: 3.4.2
 - Browser: Chrome 83
 - OS: MacOS

### Expected Behavior

Data of all forms

### Actual Behavior

I get data of last form instead of data of each form.

### Example

https://jsfiddle.net/fdehanne/1j9c8fry/14/

```
<button id="save">Save</button>
<div class="render"></div>
<div id="formEditor-fr_FR" data-culture="fr_FR" class="form-editor"></div>
<div id="formEditor-en_GB" data-culture="en_GB" class="form-editor"></div>
```

```
var formsBuilder = [];

$(function() {
  $('.form-editor').each(function() {
    var culture = $(this).attr('data-culture');
    formsBuilder[culture] = $(this).formBuilder();
  });

  $(document).on('click', '#save', function() {
    $('.render').html('');
    for (var culture in formsBuilder)
    {
      $('.render').append(culture + '<br>');
      var data = formsBuilder[culture].formData;
      $('.render').append(data + '<br>');
    }
  });
});
```

ngjulia:
I have the same issue, however I am trying to render multiple forms on one page, and formRender.('userData') only retrieves the user data from the last form as well.

myowinthein:
Only initiate multiple form builder instance with promise & recursive function. Check example here.

https://github.com/kevinchappell/formBuilder/issues/965#issuecomment-526204831

ngjulia:
Will this work for form-rendering as well?

myowinthein:
Sure, here is the sample of my usage.

### Initiate Multiple Instances
```javascript
let formIDs = ['#form1', '#form2']
let fbInstances = []
let options = [............]

let init = function(i) {
    if (i < formIDs.length) {
        options['formData'] = [] // add existing form components here (for rendering)
        $(formIDs[i]).formBuilder(options).promise.then(res => {
            fbInstances.push(res)
            i++
            init(i)
        })
    }
}

init(0)
```

> Without recursive function, all previous form instance will be override by later one. In above case, outputting of `form1` will always show as `form2`.

### Get Data from All Instances on Submit

```javascript
const formData = []

fbInstances.forEach(fbInstance => {
    formData.push(fbInstance.actions.getData())
})
```
[...]

Issue #1000, opened 2019-10-22
Duplicates data in multi-page form


Ravi-kavya01:
I am getting the last tab data for each one.

You can check this URL. It has the same issue
https://codepen.io/anon/pen/QaKrXy

Video: https://drive.google.com/file/d/1XxvHQkyRUwAEko0uXNwXszhq2ktiUIiS/view?usp=sharing

Putt the content on the last one and click save will console the data

Thanks

Ravi-kavya01:
Can any one solve this issue?

xaroze:
> I am getting the last tab data for each one.
> 
> You can check this URL. It has the same issue
> https://codepen.io/anon/pen/QaKrXy
> 
> Video: https://drive.google.com/file/d/1XxvHQkyRUwAEko0uXNwXszhq2ktiUIiS/view?usp=sharing
> 
> Putt the content on the last one and click save will console the data
> 
> Thanks

Is this issue solved, @Ravi-kavya01 ?

W1-PopelierE:
@sar0z It doesn't look like this issue has been resolved yet.

sar0z:
> @sar0z It doesn't look like this issue has been resolved yet.

Do we have any solution,  its hampering the created forms.

PolarCoffee:
any solution yet?, having the same issue

krebbi:
Still not resolved. 

@kevinchappell you can see the issue on your codepen https://codepen.io/kevinchappell/pen/BKYPRp when opening the console, create multiple tabs with different elements and hit "save all".
you get as many array elements as tabs but all to have the same content/elements as the last tab.

krebbi:
I hope this helps:

https://codepen.io/krebbi/pen/VwrPjZV?editors=1111

lucasnetau:
You need to await the promise returned by formBuilder when using multiple instances

https://formbuilder.online/docs/formBuilder/promise/

See https://github.com/kevinchappell/formBuilder/issues/1087 https://github.com/kevinchappell/formBuilder/issues/1340 https://github.com/kevinchappell/formBuilder/issues/1383

Issue #1383, opened 2023-07-27
Uncaught TypeError: fb.actions.getData is not a function


Repechinskyi:
Hello. I can't solve the problem for several days now. When I create a multi form, I add pages, fields. I will keep everything. But when I want to change this form and have more than 2 pages, I get an error on the second element "Uncaught TypeError: fb.actions.getData is not a function".

I think that the problem is in the class instance for 2 elements, but the same situation when, for example, 6 elements, 1 and the last element work fine, but all the others do not - an error

But this functionality loads the form, that is, I see all the saved fields after creation, everything is fine. but I can't save. I tried to add/remove fields all exactly error. The error disappears when I remove the intermediate pages of the form.

//--------------------

    fbInstances.push($(".build-wrap").formBuilder(fbOptions));

    $(document).ready(function() {
        setTimeout(function() {
            $.each(setFormData, function(i, item) {

                if(fbInstances[i]) {
                    fbInstances[i].actions.setData(item);
                } else {
                    addPageLoading(item);
                }

            });
        }, 2000);
    });

  /***/
  
    function addPageLoading(data){
        if(data.length){
            fbOptions.formData = data;
        }else{
            fbOptions.formData = [];
        }
        fbInstances.push($(newPage).formBuilder(fbOptions));
    }
    
    
        $(document.getElementById("getJSON")).click(function () {

        const allData = fbInstances.map((fb) => {
            var json = fb.actions.getData('json', true);
            return json;
        });
        $("input[name='json']").val("["+allData+"]");
        $("#design-form").submit();
   
    });

Repechinskyi:
@lucasnetau 

Hello. Thank you for giving a hint, which helped partially, but now I have encountered another problem in the form, the fields are displayed correctly, but when I want to get data from the form in json format, I have the same code on page 2 and 3. I've been struggling with this code for several days, I don't know what to do anymore.

I have 3 pages now:
1 - Text field
2 - Date field
3 - Header fields

These fields in the form will visually display correctly. There is no error with "actions.getData" now, but when I view the result var json = fb.actions.getData('json', true); then I get

1 - Text field
2 - Header fields
3 - Header fields

After the first it is duplicated


`document.addEventListener('DOMContentLoaded', async function() {
	jQuery( function ($) {

		 var fbOptions = {...};

		 fbInstances.push($(".build-wrap").formBuilder(fbOptions));

	     setTimeout(function() {
	        $.each(setFormData, function(i, item) {

[...]

Issue #1340, opened 2022-10-28
Multiple Formbuilder Forms on a single page not working properly.


Sharad2390:
### Environment Details: I am working on a staging environment having PHP on the stack side

 - formBuilder Version: Latest one
 - Browser: Chrome
 - OS: Windows

### Steps to Reproduce: Below is complete code which I am working on:
**JS** 
```
jQuery(function($) {
  var fbTemplate = document.getElementById('fb-editor');
  var options = {
    disabledActionButtons: ['data','clear','save']
  };
  var formBuilder = $(fbTemplate).formBuilder(options);
  document.getElementById('getJSON').addEventListener('click', function() {
    alert(formBuilder.actions.getData('json'));
  });
  var fbTemplate1 = document.getElementById('fb-editor1');
  var options1 = {
    disabledActionButtons: ['data','clear','save']
  };
  var formBuilder2 = $(fbTemplate1).formBuilder(options1);
  document.getElementById('getJSON1').addEventListener('click', function() {
    alert(formBuilder2.actions.getData('json'));
  });
});

```
**HTML**
```
<div id="fb-editor"></div>
<button id="getJSON" type="button">Get JSON Data</button>
<div id="fb-editor1"></div>
<button id="getJSON1" type="button">Get JSON Data</button>
```

Sharad2390:
Hi @kevinchappell,
Please help me with this as I am totally stuck with this issue.
Thanks!

lucasnetau:
Take a look at https://formbuilder.online/docs/formBuilder/promise/ and use the result of the initialisation promise

kirykr:
I can see this in [Section](https://formbuilder.online/docs/formBuilder/actions/getData/). But when I use promise and store the objects in an array, all the other objects get overridden by the last object.

```coffee
  @formBuilder = []
  for element in $('#tab1, #tab2')
      dataElement = JSON.parse($(element).children('span').text())
      _initProgramBuilder(element, (dataElement || []))
  
  _initProgramBuilder = (element, data) ->
    $(element).formBuilder().promise.then (fb) ->
      @formBuilder.push fb
   
```

Abacaxi-Nelson:
Same here @kirykr 
Did you find a solution, i have two form builder, and calling "actions.getData()" only return last one..
I am already using promises...

Repechinskyi:
Hello. Who was able to solve this problem? I ran into it and all methods do not help

lucasnetau:
[...]

PR #1459, opened 2023-10-19
Refactor initialisation of the formBuilder plugin to ensure two or more concurrent initialisations cannot interfere with each other

lucasnetau:
Initialising two formBuilders concurrently would cause all but the last instance to be incorrectly initialised. Refactor this to using a initialisation function to isolate the plugin instance and store it in a data attribute for later use.

Additionally an uninitialised placeholder function has been put in place when actions are called before the formBuilder is initialised (when not using promise interface)

Both of these changes should reduce issue spam and also improve new developer experience by not having to load multi formBuilders in a recursive function or awaiting in a loop.

Tests have been added to check for concurrent loading, all existing tests pass after this change.

Closes: https://github.com/kevinchappell/formBuilder/issues/1087
Closes: https://github.com/kevinchappell/formBuilder/issues/1000
References: https://github.com/kevinchappell/formBuilder/issues/965
References: https://github.com/kevinchappell/formBuilder/issues/1088
References: https://github.com/kevinchappell/formBuilder/issues/1340
References: https://github.com/kevinchappell/formBuilder/issues/1383

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -2489,68 +2489,72 @@ function FormBuilder(opts, element, $) {
   return formBuilder
 }
 
-const methods = {
-  init: (options, elems) => {
-    const { i18n, ...opts } = jQuery.extend({}, defaultOptions, options, true)
-    config.opts = opts
-    const i18nOpts = jQuery.extend({}, defaultI18n, i18n, true)
-    methods.instance = {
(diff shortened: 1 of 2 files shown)


kevinchappell (approved):
excellent change. should eliminate one of the most common issues faced when using formBuilder

PR #1463, opened 2023-10-20
Allow string row identifiers in bootstrap rows

lucasnetau:
Allow Bootstrap row identifiers to be any string value after row)- (eg row-myrow). This allows default fields or inputSets to be pre-configured with named rows which can then be targeted with CSS rules.


lucasnetau:
This also opens up future support for naming rows via the stage

kevinchappell (approved):

Issue #1461, opened 2023-10-19
Custom Control not working with Enhanced Bootstrap Grid


kguillottffbla:
### Description:
Trying to add a custom select control that is populated with an ajax query. It works fine when the enhanced bootstrap grid is not enabled. When enabled, I'm getting this error: Uncaught (in promise) TypeError: e.querySelectorAll is not a function.

```
var options = {
    enableEnhancedBootstrapGrid: true,
    showActionButtons: false,
    dataType: 'json',
    fields: [
        {
            label: 'User List',
            type: 'userlist',
            icon: '<i class="bi bi-person"></i>'
        }
    ],
    templates: {
        userlist: function (fieldData) {
            return {
                field: '<select id="' + fieldData.name + '" name="' + fieldData.name + '" class="form-select" style="width:auto"></select>',
                onRender: function () {
                    console.log(fieldData);
                    $.get('/Forms/GetActiveUsers', function (data) {
                        var selectlist = $('#' + fieldData.name);
                        $.each(data, function (index, item) {
                            var option = '<option value="' + item.userId + '">' + item.displayName + '</option>';
                            selectlist.append(option);
                        });
                    });
                }
            };
        }
    }
};

let formBuilder = $(document.getElementById('fb-editor')).formBuilder(options);
```
### Environment Details:

 - formBuilder Version: 3.16.4
 - Browser: Edge
 - OS: Windows 10

lucasnetau:
hi @kguillottffbla, the value of field is meant to be of type HTMLElement|HTMLElement[] whereas in your code (and some of the examples in the docs) returns a string. Strings work fine normally but when adding row- and col- classes this breaks down when it tried to strip the bootstrap classes from the field.

Quickest fix for you is to use:

```javascript
field: this.markup('select', null., { name:  fieldData.name, className: 'form-select', style: 'width:auto' }),
```

I'll also look at wether it makes sense to coerce the field as a string to a HTMLElement in layout.build()

lucasnetau:
@kguillottffbla 

PR https://github.com/kevinchappell/formBuilder/pull/1466 should resolve your issue

PR #1466, opened 2023-10-24
Fix processing of fields defined with a raw HTML string when combined with bootstrap row/col

lucasnetau:
Many examples for custom field templates in the documentation use simple HTML strings as the return value rather than HTMLElement(s) used by built-in controls. layout.processClassName expects the field to be HTMLElements in order to strip any bootstrap related classes from the classList.

We fix this by building the HTMLElement(s) via markup() prior to passing to the layout template

Fixes: #1461

### DIFF
--- a/src/js/layout.js
+++ b/src/js/layout.js
@@ -121,6 +121,16 @@ export default class layout {
       field = { field: field }
     }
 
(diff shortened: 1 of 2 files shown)


kevinchappell (approved):
good fix

PR #1470, opened 2023-10-25
Fix select placeholder configuration

lucasnetau:
When creating a select field with both a placeholder we set the value to an empty string '' otherwise it will take the value of the option's textContent. If value is not set to an empty string then an check for required validity will return true.

PR also updates disabled and selected to boolean values so they are rendered correctly.

Prev: `<option disabled="null" selected="null">Select an option</option>`

Now: `<option disabled="disabled" selected="selected" value>Select an option</option>`


kevinchappell (approved):
:+1:

PR #1472, opened 2023-10-26
Fix bugs related to textarea.quill

### LINKED ISSUE #677: quill editor looks wrong
### Description:
probably some css it's missing ?

### Environment Details:

 - formBuilder Version: 
 - Browser: 
 - OS: 

### Steps to Reproduce
1. add a textarea and select type quill
2. add more than one rows of text

### Screenshot - *(optional)*
quill

lucasnetau:
1. quill.textarea contains SVG graphics. This caused getAllGridRelatedClasses to throw a exception due to SVG className property being a SVGAnimatedString instead of a string
2. quill.textarea is incompatible with form-control class, strip this in the build stage Fixes #677
3. Double clicking on a textarea or on content within a content editable field (eg textarea.quill) would cause the edit box to toggle. Add textarea to the exclusion list and also use the isContentEditable property instead of the contentEditable attribute to detect the editable state

### DIFF
--- a/src/js/control/textarea.quill.js
+++ b/src/js/control/textarea.quill.js
@@ -49,6 +49,9 @@ export default class controlQuill extends controlTextarea {
     //Textareas do not have an attribute 'type'
     delete attrs['type']
     this.field = this.markup('div', null, attrs)
+    if (this.field.classList.contains('form-control')) {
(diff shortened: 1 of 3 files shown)


kevinchappell (approved):
good fixes :raised_hands:

PR #1474, opened 2023-10-31
Tests for checkboxes + dom fix

lucasnetau:
Add in test cases for the checkbox group control

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -215,12 +215,8 @@ export default class controlSelect extends control {
           }
 
           // Did not find a match for the selectedOption, see if this is an "other"
-          if (input.id.endsWith('-other')) {
-            const otherVal = document.getElementById(`${input.id}-value`)
-            // If there is no value to set, don't check the other option
-            if (selectedOptions.length === 0) {
-              return
-            }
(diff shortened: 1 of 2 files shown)


src/js/control/select.js
```diff
@@ -215,12 +215,8 @@ export default class controlSelect extends control {
...
 
           // Did not find a match for the selectedOption, see if this is an "other"
-          if (input.id.endsWith('-other')) {
-            const otherVal = document.getElementById(`${input.id}-value`)
-            // If there is no value to set, don't check the other option
-            if (selectedOptions.length === 0) {
-              return
-            }
+          if (input.id.endsWith('-other') && selectedOptions.length > 0) {
+            const otherVal = this.dom.querySelector(`#${input.id}-value`)
```
kevinchappell, line 219: just curious, is this a better way of selecting ids?

kevinchappell (approved):
good test :+1:

src/js/control/select.js
lucasnetau: getElementById is only available on document. When selecting ids as children of an element you need to use the query selector. Because there is no guarantee that the control is part of the document yet (formRender wrapped not attached) we need to do it this way,.

Issue #995, opened 2019-09-30
templates API fieldData properties loss


Zulkifilyusuf750:
### Description:
```
let templates = {
  'gps-coords': function(fieldData) {
    console.log(fieldData);
    };
  }
};
```
The **fieldData**.required, fD.title, fD.description, fD.className properties get loss, and are always undefined in a template (or custom type).
### Environment Details:

 - formBuilder Version: 3.2.5
 - Browser: Chrome
 - OS: Windows 10

### Expected Behavior
fieldData properties to be set on field add & field edit.
### Actual Behavior
Some properties of fieldData are always undefined (only set in rawConfig property).
### Screenshot - *(optional)*
![2019-09-14 20_18_27-Greenshot](https://user-images.githubusercontent.com/20576428/65901203-9def5300-e3af-11e9-8440-d37fffcdb574.png)

lucasnetau:
Currently these are the properties that a custom control are given access to. title and description are part of the list 

https://github.com/kevinchappell/formBuilder/blob/27abe220d4dabe8d08a2ab64c28d8e98a986854f/src/js/control/custom.js#L125-L135

emirdeliz:
@lucasnetau I have the same problem here. 

```javascript
const customFormTemplates = {
  custom: function (fieldData) {

    console.log({ fieldData });
  
    const { placeholder, className, name, fieldType } = fieldData;
    return {
      field: `
        <input
          type="text"
          class="form-control ${className || ''}"
          placeholder="${placeholder || ''}"
          name="${name || ''}"
          id="${fieldData.name}"
          fieldType="${fieldType}"
          readonly
        />
      `,
    };
  },
```

The console.log always displays undefined for the required.

lucasnetau:
@emirdeliz, this is as designed (as far as I can tell going through the code) required is removed in formBuilder when rendering the preview to stop the required attribute being set on the inputs, however the preview will still show the required asterisk * and formRender will still render the custom template correctly

https://github.com/kevinchappell/formBuilder/blob/64dfde5961d867d8ac4156eb967b95b3af94d675/src/js/control.js#L23-L25

The only issue I've found so far is that isPreview is in the fieldData object but is always empty because of this line
https://github.com/kevinchappell/formBuilder/blob/64dfde5961d867d8ac4156eb967b95b3af94d675/src/js/control.js#L22

and is therefore not able to be set in custom.js because it doesn't exist. 
[...]

PR #1475, opened 2023-10-31
Set the preview property for custom controls

lucasnetau:
Custom control was attempting to use the property isPreview, this does not exist as it is deleted in control.js constructor. Set the preview property in the same way we send along preview to built-in controls

Fixes: #995


kevinchappell (approved):
awesome. good to get tests around custom controls

Issue #1477, opened 2023-10-31
userData is undefined for checkbox-group if no options selected, should be empty array


vijay-torani:
### Description:
check box group is not working when i unchecked and get userdata it still show checked in userdata.
### Environment Details:

 - formBuilder Version: 
  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.1/jquery.min.js"></script>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.13.2/jquery-ui.min.js"></script>
  <script src="https://formbuilder.online/assets/js/form-builder.min.js"></script>
 - Browser: chrome
 - OS: windows

### Expected Behavior
Should Be Unchecked if i Unchecked the checkbox
### Actual Behavior
it remain unchecked if i unchecked 
### Screenshot - *(optional)*
![Screenshot 2023-10-31 113931](https://github.com/kevinchappell/formBuilder/assets/107983453/e7696561-bc93-422c-b2c9-97e57b5e02cc)
![Screenshot 2023-10-31 114017](https://github.com/kevinchappell/formBuilder/assets/107983453/3abb272d-20da-45a6-a647-7831458fb760)
![Screenshot 2023-10-31 114038](https://github.com/kevinchappell/formBuilder/assets/107983453/8708c2a8-4517-4fca-bcf4-24e4e1d4f39b)

lucasnetau:
The issue here is that there is no userData being returned, if you try to reload the form then it will use the default selected options rather than unsetting them

lucasnetau:
$('select, input, textarea', container).serializeArray() will remove the input[type=checkbox] if none selected

vijay-torani:
> The issue here is that there is no userData being returned, if you try to reload the form then it will use the default selected options rather than unsetting them

but i am console the data in console data you Cleary see i am getting checkbox checked but i unchecked the checkbox

lucasnetau:
No. Your console output had no userData key for the checklist. The 'selected' value you are seeing is the option values set by formBuilder, they are not changed by formRender

vijay-torani:
that't the point why i am not getting user data for checkbox

lucasnetau:
Please read above. It was a bug which was fixed in a release today. Please retest

vijay-torani:
ok thank you sir

PR #1478, opened 2023-10-31
Fix GH-1477 userData incorrect when unchecking defaulted selected checkboxes

lucasnetau:
@kevinchappell I believe setting this default empty array is the best method to ensure than when no value is available it is set. This ensures that any default values are not selected on subsequent loads.

Fixes: #1477


kevinchappell:
> @kevinchappell I believe setting this default empty array is the best method to ensure than when no value is available it is set. This ensures that any default values are not selected on subsequent loads.
> 
> Fixes: #1477

I agree, that is a good solution

Issue #1480, opened 2023-11-02
control.controlConfig shared between instances


lucasnetau:
### Description:

control.controlConfig static is global not per instance so subsequent loads of formBuilder will overwrite previous controlConfig data

### Environment Details:

 - formBuilder Version: Current
 - Browser: Any
 - OS: Any

### Expected Behavior

controlConfig is isolated per formBuilder instance

### Actual Behavior

controlConfig is global and can be overwritten by subsequent loads

PR #1481, opened 2023-11-03
Fix GH-1480 ensure that controlConfig is loaded per formBuilder instance.

lucasnetau:
control.controlConfig static property is global not per instance so it would be overwritten by subsequent loads of formBuilder. Fix this in a BC way (keeping static property) for any custom controls out there and load the config at the time of constructing the control class

Fixes: #1480

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -21,9 +21,6 @@ export default class Controls {
     this.opts = opts
     this.dom = d.controls
     this.getRegistered = control.getRegistered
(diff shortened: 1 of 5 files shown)


lucasnetau:
@kevinchappell I can't think of a cleaner way to fix this while maintaining BC compat with any custom plugins. There are a few more places where there are global static class properties that are shared between instances of formBuilder however I'm holding off rewriting those modules right now.

PR #1482, opened 2023-11-05
Bootstrap feature updates

lucasnetau:
- Remove redundant checkSetupBlankStage() check and row creation. Changes in previous PRs made this redundant and the original code had an issue leaving a dangling row in the stage.
- Remove a redundant branch between dropping to a new row or dropping to a placeholder
- Add additional checkRowCleanup() after dropping controls. Mouse drag of controls could sometimes lead to placeholders not been cleanup and placing whitespace through the stage.
- Store the row ID in a data-row-id attribute. This allows CSS rules to display information like row name in a information panel above each row.
- Fallback to showing controlType: name in formBuilder when the label is blank. This improves UX of controls on the stage

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1341,7 +1341,6 @@ function FormBuilder(opts, element, $) {
 
             setupSortableRowWrapper(rowWrapperNode)
             syncFieldWithNewRow(colWrapper.attr('id'))
-            checkRowCleanup()
           }
         }
 
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -1846,7 +1846,6 @@ function FormBuilder(opts, element, $) {
   if (enhancedBootstrapEnabled()) {
     $stage.on('stageEmptied', () => {
       formRows = [] //Reset row count
-      checkSetupBlankStage()
     })
   }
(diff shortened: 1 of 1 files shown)


src/js/form-builder.js
```diff
@@ -1316,34 +1317,32 @@ function FormBuilder(opts, element, $) {
...
-        }
+            colWrapper.appendTo(rowWrapperNode)
 
-        if (droppingToPlaceholderRow && senderIsControlsBox) {
-          insertTargetIsRow = true
-          insertingNewControl = true
-          $targetInsertWrapper = $(ui.item).parent()
+            setupSortableRowWrapper(rowWrapperNode)
+            syncFieldWithNewRow(colWrapper.attr('id'))
+            checkRowCleanup()
```
kevinchappell, line 1344: do we need this if its called below?

kevinchappell (approved):

kevinchappell:
just had the one question but this is good to merge when youre ready

src/js/form-builder.js
lucasnetau: You are correct. I've removed the redundant checkRowCleanup() here

Issue #1484, opened 2023-11-17
Text Area Subtype not being disabled


isakyaki28:
Hi, so I've added the subtype field to be disabled, in hopes of removing the Type field in the Text Area settings, however it's still showing up.
` disabledAttrs: ["access", "className", "name", "maxlength", "rows", "subtype", "toggle", "value"]`
Also tried using the disabledSubtypes, as I only want the Text Area option and not the other two, but those two options are still appearing.
` disabledSubtypes: ["quill", "tinymce"]`

### Screenshot - *(optional)*
<img width="968" alt="Screenshot 2023-11-17 at 12 14 49 PM" src="https://github.com/kevinchappell/formBuilder/assets/147768763/1b625291-dd64-4d8f-9e0e-6428f2316ff6">
<img width="832" alt="Screenshot 2023-11-17 at 12 15 18 PM" src="https://github.com/kevinchappell/formBuilder/assets/147768763/967ddd7d-0279-4af0-9c1f-a60d1622f644">

Would appreciate any help!

lucasnetau:
Hi @isakyaki28 ,

disabledSubtypes needs to type to be defined, please use this

`disabledSubtypes: { textarea: ["quill", "tinymce"] },`

There is a visibility bug where subtype should be hidden if disabled (You cannot disable subtype only disable otherwise some fields will fail)

`subtype: () => selectAttribute('subtype', values, subtypes[type]),` needs to take into account isHidden, will prep a PR

isakyaki28:
thanks @lucasnetau! adding the textarea for disabling the subtype works.

PR #1486, opened 2023-11-22
Ensure subtype attribute is hidden when disabled via disableAttr

lucasnetau:
subtype was recently added as a special attribute that gets hidden rather than removed when disabled (otherwise fields will break). This isHidden flag was not supported/propagated to the selectAttribute and as such did not hide the subtype attribute.

Tests added for disableAttr option

Fixes: #1484


kevinchappell (approved):
nice change

Issue #699, opened 2018-02-02
Value attribute for TextArea control is a input[type=text] and not textarea/div[contenteditable] leading to removal of new lines


sergioralmeida:
We can write new lines (hit the "enter" button) on the textarea, but it doesn't save them.
You can try this on the demo page.
1 - https://formbuilder.online/
2 - Add a textarea
3 - Write something with multiple lines
4 - Save
5 - Weep

![image](https://user-images.githubusercontent.com/5584428/35748192-0bd22d5e-0845-11e8-9302-a3ee379eea5f.png)

Value is saved with all lines together, not even a tiny space between characters.

![image](https://user-images.githubusercontent.com/5584428/35748215-20c83050-0845-11e8-9fc4-53026f3e3c08.png)

Any ideas to fix this are highly appreciated. Cheers.

YaroslavPetryk:
you need at first before sending to encodeURIcomponent(message), then when you recieve the message decodeURIComponent(message); then message.replace(/%0A/g, "<br>");

Jojoshua:
Is this still an issue?

lucasnetau:
The value field is a a input[type=text] element which can only take single line text. So when we sync the value of the preview textarea to the value field the \n newlines are removed.

https://github.com/kevinchappell/formBuilder/blob/58baf4e4de05f98406ea7050a12ae87cef5a1140/src/js/form-builder.js#L1717

temamagic:
Also in edit mode, we have single line input instead textarea
<img width="231" alt="image" src="https://github.com/kevinchappell/formBuilder/assets/15264697/f3f862ba-6823-4be1-811f-d2f0b960bf35">

PR #1487, opened 2023-11-23
Fix for https://github.com/kevinchappell/formBuilder/issues/699

Shinbar:
After recently using this great project, I found that TextArea content are not being preserved, as they are being converted into a single value input of input[type=text].  This removes formatting from the TextArea such as line feeds etc.

I have amended the code so that TextArea are provided a proper TextArea control, and that the innerHTML data is preserved.

I believe this solved the following [issue](https://github.com/kevinchappell/formBuilder/issues/699).

Fixes: #699

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -936,7 +936,6 @@ function FormBuilder(opts, element, $) {
 
     let attrVal = values[attribute] || ''
     let attrLabel = mi18n.get(attribute) || attribute
-    const typeVal = values['type'] || ''
 
     if (attribute === 'label') {
       if (textArea.includes(values.type)) {
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -964,10 +964,8 @@ function FormBuilder(opts, element, $) {
       if (attribute === 'label' && !opts.disableHTMLLabels) {
         inputConfig.contenteditable = true
         attributefield += m('div', attrVal, inputConfig).outerHTML
-      } else if (typeVal.includes('textarea') && attribute === 'value') {
-        inputConfig.contenteditable = true
-        inputConfig.value = attrVal
-        attributefield += `<textarea ${attrString(inputConfig)}>${inputConfig.value}</textarea>`
+      } else if (values.type === 'textarea' && attribute === 'value') {
+        attributefield += m('textarea', attrVal, inputConfig).outerHTML
       } else {
         inputConfig.value = attrVal
         inputConfig.type = 'text'
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -936,6 +936,7 @@ function FormBuilder(opts, element, $) {
 
     let attrVal = values[attribute] || ''
     let attrLabel = mi18n.get(attribute) || attribute
+    const typeVal = values['type'] || ''
 
     if (attribute === 'label') {
       if (textArea.includes(values.type)) {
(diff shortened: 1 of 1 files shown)


lucasnetau:
Thank you @Shinbar for your PR. I'll will review

src/js/form-builder.js
```diff
@@ -963,6 +964,10 @@ function FormBuilder(opts, element, $) {
       if (attribute === 'label' && !opts.disableHTMLLabels) {
         inputConfig.contenteditable = true
         attributefield += m('div', attrVal, inputConfig).outerHTML
+      } else if (typeVal.includes('textarea') && attribute === 'value') {
+        inputConfig.contenteditable = true
+        inputConfig.value = attrVal
+        attributefield += `<textarea ${attrString(inputConfig)}>${inputConfig.value}</textarea>`
```
lucasnetau, line 970:
```suggestion
      } else if (values.type === 'textarea' && attribute === 'value') {
        attributefield += m('textarea', attrVal, inputConfig).outerHTML
```

textarea doesn't need the contenteditable set and we can use the markup helper rather than manually constructing the HTML element

src/js/form-builder.js
```diff
@@ -1713,6 +1718,11 @@ function FormBuilder(opts, element, $) {
           selectedOption.checked = prevOptions[i].checked
         })
       }
+    } else if (field.type === 'textarea') {
+      const fieldVal = document.getElementById('value-' + field.id)
+      if (fieldVal) {
+        fieldVal.innerHTML = e.target.value
+      }
```
lucasnetau, line 1725:
```suggestion
```

Why have you chosen to convert between raw string value and HTML? We can pass the value directly between the value get/set for the elements `fieldVal.value = e.target.value`

src/js/form-builder.js
```diff
@@ -936,6 +936,7 @@ function FormBuilder(opts, element, $) {
 
     let attrVal = values[attribute] || ''
     let attrLabel = mi18n.get(attribute) || attribute
+    const typeVal = values['type'] || ''
```
lucasnetau, line 939:
```suggestion
```

I don't think you need to define an extra variable here. The type can never be undefined nor and we only reference it once in the code below.

lucasnetau (changes requested):
Hi @Shinbar thanks for the PR. I've added some comments and suggested changes.

lucasnetau:
Also, if possible can you please adjust your commit message to include the prefix 'fix: ' so it becomes


fix: For TextArea control have made the value input a textarea and not a single input[type=text] value. Formatting of TextAreas is now preserved.

src/js/form-builder.js
Shinbar: Your suggested change looks good to me.  I was a little unhappy with the backtick manual html.  Was just following the lines below this code which sets up a single input line 975, but using the m function is the way forward.

src/js/form-builder.js
Shinbar: In the previous set I was using the contenteditable which meant that the helper.js function getAttrVals function was using the innerHTML.  Since you amendment removed that I agree no point to this piece of code anymore.

src/js/form-builder.js
Shinbar: Yup it left over from the previous code amendments.  No longer required.

lucasnetau:
LGTM

@kevinchappell any issues with this change for textareas?

kevinchappell (approved):
changes look good. thank you @Shinbar

PR #1489, opened 2023-11-28
Multiple Bootstrap improvements

lucasnetau:
- Only setup sortable row wrapper and invisible row placeholders when adding a row to the stage. 
- Calling SetupInvisibleRowPlaceholders with an existing row led to duplicated invisible row placeholders.
- Calling setupSortableRowWrapper on an existing row causes unnecessary reinitialisation of jQuery.ui sortable
- When initialising from formBuilder don't move the row to the LI insertion point since we will be moving the LI into the row in the next line.
- Fix some jumping around of row placeholders due to control panel sortable configuration using intersect rather than pointer


kevinchappell (approved):
good work!

PR #1515, opened 2024-02-08
Fix formRender removing a valid empty userData array

### LINKED ISSUE #1514: checkbox is auto seleced when dragged and dropped, should be deselected initially when dragged and dropped
checkbox is auto seleced when dragged and dropped, should be deselected initially when dragged and dropped

lucasnetau:
select.js controls (Radio, Checkbox, Select) support an empty userData array. This was being filtered out by sanitizeField when putting the field data through trimObj.

This PR:

- fixes up the documentation for trimObj to document the removal of empty arrays
- ensures an empty userData array is returned while allowing trimObj to remove other invalid datatypes
- fixes a incorrect assignment when sanitising the value object
- adds additional tests to validate this behaviour

Fixes #1514

### DIFF
--- a/src/js/form-render.js
+++ b/src/js/form-render.js
@@ -156,7 +156,7 @@ class FormRender {
       : field.className || field.class || null
     delete sanitizedField.class
     if (field.values) {
-      field.values = field.values.map(option => utils.trimObj(option))
+      sanitizedField.values = field.values.map(option => utils.trimObj(option))
     }
     sanitizedField = utils.trimObj(sanitizedField)
     if (Array.isArray(field.userData) && field.userData.length === 0) {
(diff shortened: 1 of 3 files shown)


src/js/form-render.js
```diff
@@ -156,9 +156,13 @@ class FormRender {
...
     delete sanitizedField.class
     if (field.values) {
-      field.values = field.values.map(option => utils.trimObj(option))
+      sanitizedField.values = field.values.map(option => utils.trimObj(option))
     }
-    return utils.trimObj(sanitizedField)
+    sanitizedField = utils.trimObj(sanitizedField)
+    if (Array.isArray(field.userData) && field.userData.length === 0) {
+      sanitizedField.userData = [] //Special handler for allowing userData to be empty
+    }
```
kevinchappell, line 164: :+1:

kevinchappell (approved):

PR #1521, opened 2024-02-19
Fix processing of number attributes for a field

### LINKED ISSUE #1520: Default value for number attribute is not used
### Environment Details:

 - formBuilder Version: Any
 - Browser:  Any
 - OS:  Any 

### Expected Behavior

When using

```javascript
typeUserAttrs: {
        text: {
          testAttribute: {
            label: 'test',
            value: 10,
          },
        },
      },
```
and adding a text box the attribute test should be set to 10

### Actual Behaviour

Value is undefined

lucasnetau:
numberAttribute function loads the value from attributes[attributeName], this was not set for userTypeAttrs. Assign the value into the tUA array 

Additional tests added around adding fields with number attributes to the stage 

Fixes #1520

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -33,7 +33,7 @@ import {
   safename,
   forceNumber,
   getContentType,
-  generateSelectorClassNames,
+  generateSelectorClassNames, firstNumberOrUndefined,
 } from './utils'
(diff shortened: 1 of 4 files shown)


kevinchappell (approved):
good work :+1:

PR #1522, opened 2024-02-20
Ensure config is per instance

lucasnetau:
Previously config was a global const shared between formBuilder instances. Defaults and config set in the last instance would override all previous instance. For the most this was not noticed because the  per instance opts object is passed to controls but helpers and some callbacks like onAddOption were accessed via the global const config.

All existing tests pass.


kevinchappell (approved):
huge win for multiple instances with different configs :tada:

PR #1526, opened 2024-02-26
Improve stage and control layout with flexbox

### LINKED ISSUE #1524: If i give large Text for header component in my form builder, UI is getting disrupted. There is no way to limit the length as header component does not p
### Environment Details:

 - formBuilder Version: 
 - Browser: 
 - OS: 

### Screenshot - *(optional)*
headdeerrr

lucasnetau:
* Stage and control layout moved from floats to flexbox
* Overflow-wrap implements for text to ensure encapsulation in the stage Fixes #1524
* Remove $(window).scroll handler with flex sticky CSS

### DIFF
--- a/src/js/data.js
+++ b/src/js/data.js
@@ -13,7 +13,6 @@ export class Data {
   constructor(formID) {
     this.formData = {}
     this.formID = formID
-    this.layout = ''
     instanceData[formID] = this
   }
 }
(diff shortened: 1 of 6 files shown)


kevinchappell (approved):
awesome. originally the sticky controls was using javascript to ensure the controls were sticky to the parent container and i think handle some changes on resize. i'm glad this can be done with pure css now!

Issue #1530, opened 2024-03-03
Issue setData


Fharhanamrin:
### Environment Details:

 - formBuilder Version: 
 - Browser: 
<img width="1166" alt="Screenshot 2024-03-04 at 00 11 51" src="https://github.com/kevinchappell/formBuilder/assets/27546921/aec0748f-fafc-4720-9f86-634014b375ac">

 - OS: 

### Steps to Reproduce
Insert Form => set Data form =>  does not match what was saved
### Screenshot - *(optional)*
insert to db
<img width="974" alt="Screenshot 2024-03-04 at 00 14 28" src="https://github.com/kevinchappell/formBuilder/assets/27546921/0c3298c1-07ad-4f65-8994-d65cafa391c3">
appears after saving the data value label indonesia and label inggris wrong
<img width="958" alt="Screenshot 2024-03-04 at 00 15 29" src="https://github.com/kevinchappell/formBuilder/assets/27546921/6cb84704-c311-4c1d-adfc-c25997077709">

```
 const fbEditor = document.getElementById("build-wrap");
    const templateSelect = document.getElementById("formTemplates");
    var fields = [{
            label: "Select",
            type: "select",
            subtype: "Select",
            icon: "✉",
            values: [{
                    label: "Label Indonesia",
                    value: "Label Inggris",
                    // selected: true
                },
                {
                    label: "",
                    value: "",
                },
                {
                    label: "",
                    value: "",
                }
            ]
        },
        {
            label: "FullName",
            type: "text",
            icon: "✉"
        }
    ];

    var options = {
        roles: {
            1: 'Form',
            2: 'Copy Data from Buyer'
        },
        replaceFields: [{
                type: "checkbox-group",
                label: "Checkbox",
                values: [{
                    label: "Label Indonesia",
                    value: "Label Inggris"
                }],
                icon: "☑"
            },
            {
                type: "radio-group",
                label: "Radio Group",
                values: [{
                    label: "Label Indonesia",
                    value: "Label Inggris"
                }],
                icon: "☑"
            }
        ],
        disableFields: [
            'autocomplete',
            'hidden',
            'number',
            'date',
            'paragraph',
            'select',
            'button'
        ],
        typeUserDisabledAttrs: {
            'text': [
                'name',
                // 'label',
                'subtype',
                // 'access',
                'helptext',
                // 'placeholder',
                'value',
                'maxlength',
                'description',
            ],
            'select': [
                'name',
                'label',
                'multiple',
[...]

PR #1531, opened 2024-03-04
Fix regression in v3.19.2, typeUserAttrs of type string failed to restore user set formData

lucasnetau:
inputUserAttrs only tried to load values from attrs[value] and not attrs[attributeName] like other typeUserAttrs helpers, this meant that only the value from the definition was used and not any value set for the attribute in the formData.

All bases covered with typeUserAttrs by setting attrs[value] to attrs[attributeName] during attribute rendering and also inputUserAttrs modified to check for the attribute value. 

Test cases added for string, number, boolean (checkbox) and array (select) typeUserAttrs

Fixes: #1530
Fixes: #1533

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -704,7 +704,6 @@ function FormBuilder(opts, element, $) {
           const tUA = typeUserAttr[attribute]
           let origValue = tUA.value
           if (attrValType === 'boolean') {
-            origValue = tUA.value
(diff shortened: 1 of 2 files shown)


isnanwicaks (approved):

Fharhanamrin:
Screenshot 2024-03-04 at 20 08 48

problem solved, thank you sir
When will it be released into production?

PR #1550, opened 2024-04-06
chore: update bootstrap grid funcitonality

### LINKED ISSUE #1549: Inline Controls
### Description:
First-time user of formBuilder here. Thanks for creating such a helpful library. Didn't see an obvious way to display controls line, meaning two or more controls placed next to each other instead of on top of each other. Wondering if it's possible. Thanks for your help.

### Environment Details:

 - formBuilder Version: latest
 - Browser: Google Chrome 121.0.6167.161
 - OS: Windows 10

### Expected Behavior
Drag and drop a control onto the page. Drag another control onto the page but drop it to the right or left of the first control. The second control now sits to the left or right of the first control.

### Actual Behavior
The second control sits above or below the first control.

kevinchappell:
Update the demo to show bootstrap grid/inline controls

resolves #1549

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -33,14 +33,14 @@ import {
   safename,
   forceNumber,
   getContentType,
-  generateSelectorClassNames, firstNumberOrUndefined,
+  generateSelectorClassNames,
+  firstNumberOrUndefined,
 } from './utils'
 import { attributeWillClobber, setElementContent, setSanitizerConfig } from './sanitizer'
 import fontConfig from '../fonts/config.json'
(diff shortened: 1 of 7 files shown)


src/js/form-builder.js
```diff
@@ -33,14 +33,14 @@ import {
   safename,
   forceNumber,
   getContentType,
-  generateSelectorClassNames, firstNumberOrUndefined,
+  generateSelectorClassNames,
```
kevinchappell, line 36: i think all auto-formatting changes in this file

PR #1562, opened 2024-05-30
Update getting-started.md

sathishreddynp:
Form render has to be include ld If we are rendering the form. But it hasn't included in docs


kevinchappell (approved):
:+1:

PR #1571, opened 2024-07-01
Fix GH-1536 kc-toggle doesn't work with duplicated ids

### LINKED ISSUE #1536: when we have two forms in the same page, the toggle checkbox doesn't work in the second form
### Description:
when we have two forms in the same page, the toggle checkbox doesn't work in the second form

### Environment Details:

 - formBuilder Version: latest
 - Browser: chrome/safari
 - OS: macosx

### Expected Behavior

checkbox to tggle

### Actual Behavior

the checkbox in the second form doesn't toggle

### Steps to Reproduce

put two forms in the same html page, both having toggle checkbox groups

### Screenshot - *(optional)*

Referencing https://github.com/kevinchappell/formBuilder/pull/24

lucasnetau:
If there is a duplicate ID for an input kc-toggle will not work for the subsequent input due to the for attribute sending the event to the first input.

Fixes #1536

### DIFF
--- a/src/js/control/select.js
+++ b/src/js/control/select.js
@@ -93,6 +93,7 @@ export default class controlSelect extends control {
           const labelAttrs = { for: optionAttrs.id }
           let output = [input, this.markup('label', labelContents, labelAttrs)]
           if (toggle) {
+            delete labelAttrs.for
             labelAttrs.className = 'kc-toggle'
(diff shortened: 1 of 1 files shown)


kevinchappell (approved):
We should consider removing or improving KC toggle. It's quite old fashioned.

PR #1576, opened 2024-07-02
Added a new signature pad element and styling for the new element

keerthi-balaji:
The signature pad element is designed using canvas and styled the element to differentiate it from its surrounding.

### DIFF
--- a/src/sass/_controls.scss
+++ b/src/sass/_controls.scss
@@ -178,3 +178,19 @@
     }
   }
 }
+.signature-container {
+  border: 2px solid #000; 
+  padding: 10px; 
+  margin-bottom: 10px; 
+  border-radius: 5px; 
+}
+
+.signature-pad {
+  //display: block;
+  width: 100%; 
+  height: 100%; 
+}
+.signature-pad.custom-style {
+  border: 2px solid #000;
+  background-color: #f9f9f9;
+}
--- a/src/sass/form-render.scss
+++ b/src/sass/form-render.scss
@@ -46,4 +46,21 @@
       height: auto;
     }
   }
+  .signature-container {
+    border: 2px solid #000; 
+    padding: 10px; 
+    margin-bottom: 10px; 
+    border-radius: 5px; 
+  }
+  
+  .signature-pad {
+    //display: block;
+    width: 100%; 
+    height: 100%; 
+  }
+  .signature-pad.custom-style {
+    border: 2px solid #000;
+    background-color: #f9f9f9;
+  }
+  
 }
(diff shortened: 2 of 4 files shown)


lucasnetau:
Thank you for your contribution, I'll give this a test run. Could you please create some Jest tests to ensure the control is working as expected?

lucasnetau:
@keerthi-balaji Has the code for the signature pad come from another project or is this an original implementation? I know of a few plugins out there https://github.com/szimek/signature_pad with advanced implementations of a canvas signatory pad.

kevinchappell (changes requested):
Excellent contribution! Signature pad is a long requested feature and I like that this is a simple implementation that is small enough to be bundled with formBuilder.

As @lucasnetau mentioned it needs some tests, here is an example of a custom control test here: https://github.com/kevinchappell/formBuilder/blob/master/tests/form-builder-custom.test.js

Some additional thoughts:
- The label needs some work when rendered
- Can the Clear button be moved to appear inside the signature pad to avoid confusing it with a form clear button?
- The `value` is not saved with other user values in FormData or the `userData`

image


Overall great addition, just needs a few minor tweaks.

keerthi-balaji:
> @keerthi-balaji Has the code for the signature pad come from another project or is this an original implementation? I know of a few plugins out there https://github.com/szimek/signature_pad with advanced implementations of a canvas signatory pad.

It is an original implementation... I was just wanting to add a signature pad element to the form builder.

keerthi-balaji:
Hi, sorry it took so long but I did make the changes as requested.

src/sass/_controls.scss
```diff
@@ -178,3 +178,26 @@
...
+.signature-pad {
+  width: 100%; 
+  height: 100%; 
+}
+.signature-pad.custom-style {
+  border: 2px solid #000;
+  background-color: #f9f9f9;
+}
+.clear-button {
+  top: 10px; /* Adjust as needed */
```
kevinchappell, line 196: :scissors: these comment, looks ai generated.

src/sass/form-render.scss
```diff
@@ -46,4 +46,28 @@
...
+.signature-pad {
+  width: 100%; 
+  height: 100%; 
+}
+.signature-pad.custom-style {
+  border: 2px solid #000;
+  background-color: #f9f9f9;
+}
+.clear-button {
+  top: 10px; /* Adjust as needed */
```
kevinchappell, line 65: this looks repeated from above. maybe signature style can be its own file that is imported where needed.

kevinchappell (approved):
looks good, will do some local testing. Thanks for addressing the concerns and adding tests! i still think we will want to move this to a plugin in the near future as signature pad doesn't seem to be a common control and a new one appear after an update might not be a desired effect.

PR #1579, opened 2024-07-04
syncFieldWithNewRow incorrectly replacing row values className attribute when element not current within row

### LINKED ISSUE #1575: How to make bootstrap classes not disappear when removing a field from a row and putting it in an empty place
### Description:
Hello, i have a question. When i put two elements on a row, and then take one of them outside of it, it loses the bootstrap classes it had by default. I have enableEnhancedBootstrapGrid and enableColumnInsertMenu as true. I would like the elements to keep the classes after the process. 

### Environment Details:

 - formBuilder Version: 3.19.7
 - Browser: Firefox
 - OS: Linux Mint

### Expected Behavior
For the elements to keep the classes when taken out of a row.

### Actual Behavior
They lose the col, row and form-control classes

lucasnetau:
Fixes #1575

When drag + drop a field from one row to a new row the function syncFieldWithNewRow() is called 3 times, once in the sortable update: callback which removes the existing row, and again in receive: callback. On the second call the oldRow value is an empty string, which resulted in the replace() call having an empty pattern and corrupting the classString.

Reimplemented syncFieldWithNewRow by operating on the classNames as an array, handling the case where we drop onto the same row or no row value exists.

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -2139,13 +2139,27 @@ function FormBuilder(opts, element, $) {
     })
   }
 
+  /**
+   * Updates the field's className to include the current wrapping row, removing the previous row if defined
+   * @param fieldID
+   */
   function syncFieldWithNewRow(fieldID) {
     if (fieldID) {
       const inputClassElement = $(`#className-${fieldID.replace('-cont', '')}`)
-      if (inputClassElement.val()) {
-        const oldRow = h.getRowClass(inputClassElement.val())
+      const currentClassName = inputClassElement.val().trim()
+      if (currentClassName) {
+        let currentClasses = currentClassName.split(' ')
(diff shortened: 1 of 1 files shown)


lucasnetau:
Original reporter confirms this resolves their issue

src/js/helpers.js
```diff
@@ -1462,4 +1462,32 @@ export default class Helpers {
...
+            currentClasses = currentClasses.filter(function(obj) {
+              return obj !== oldRow
+            })
+          }
+          if (newRow) {
+            currentClasses.push(newRow)
+          }
+          inputClassElement.val(currentClasses.join(' '))
+        }
+      }
```
kevinchappell, line 1490: :+1:

kevinchappell (approved):
looks good to me. nice fix.

PR #1580, opened 2024-07-04
Fix GH-1578 incorrect handling of label translation for custom controls

### LINKED ISSUE #1578: Labels for fields and replaceFields options do not use localized strings
### Description:
When using the fields or replaceFields option the label attribute does not use the string from the language file.
e.g.
```
jQuery(function ($) {

    var options = {
        i18n: {
            locale: 'en-US',
            location: 'languages',
       },
       replaceFields: [{
        	label: ['dateAndTimeLabel'],
		type: 'date' 
	}]
	}
        var fbEditor = document.getElementById('fb-editor');
        var formBuilder = $(fbEditor).formBuilder(options);
});
```

What I was originally tying to achieve was to rename the date control to "Date and Time" and have the drop down list say "Date". Currently, they both share the same variable e,g, "dateField" so always end up with a shared string.
e.g.
The circled strings all share the same variable
image

### Environment Details:

 - formBuilder Version: v3.19.7
 - Browser: Chrome
 - OS: Windows

### Expected Behavior
The label for the object in the replaceFields array should be taken from the language specific file e.g. en-US.lang

### Actual Behavior
The label for the object in the replaceFields array just uses the specified string literal.

### Steps to Reproduce
See code above

### Minimal sample
[datetest.html.txt](https://github.com/user-attachments/files/16086783/datetest.html.txt)
[en-US.lang.txt](https://github.com/user-attachments/files/16086785/en-US.lang.txt)

lucasnetau:
Fixes #1578

A few issues here:
* field.label wasn't translated when constructing the custom control label and icon map
* custom control label and icon functions didn't reference the correct definition object
* translated label wasn't read when constructing the control panel
* translated label wasn't read when adding a field from the control panel to the stage

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -66,8 +66,10 @@ export default class Controls {
       // first check if this is a custom control
       let custom = this.custom.lookup(type)
       let controlClass
+      let label
       if (custom) {
         controlClass = custom.class
+        label = this.custom.label(type)
       } else {
         custom = {}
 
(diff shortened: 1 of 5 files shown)


lucasnetau:
Original reported has tested PR and confirmed their issue is resolved. @kevinchappell should be good to merge

kevinchappell (approved):
Awesome work

PR #1590, opened 2024-11-13
Ensure languages are loaded before rendering forms

### LINKED ISSUE #1577: Strings set in a custom control that use custom lang file show as undefined when first rendered but work on second render
### Description:
If you create a custom control that references strings via mi18n that are set in a custom .lang file. 

All works fine in the builder and strings are set correctly, however if you then render the form the strings that are set show as undefined.
If you then render a second time strings will set correctly. 

Below test custom control in builder
Screenshot 2024-07-03 100309

Then when rendered

image

When rendered a second time show as normal. 

image

It will also work as expected if you render the form first before adding the custom control then adding the control and rendering. 

Example of the ref in code to one of the strings. 

image

And the i18n options passed to render in the main index (for this test)
image

In debugging it seemed that the html is displayed before the applyLanguage function has finished. 

### Environment Details:

 - formBuilder Version:  3.19.7
 - Browser: Chrome
 - OS:  Windows

### Expected Behavior
Strings should be applied from the custom lang file on first render.

### Actual Behavior
Strings do not get applied from the custom .land file until second render. 

### Steps to Reproduce
Create a custom .lang file and add the details to the i18n location. 
Create a custom control and add refs to any custom strings in your custom lang file using controlClass.mi18n("yourCustomStringKey").
Start Form builder and add your control to your form (All strings added as expected)
Render the form (Custom strings used show as undefined)
Render again (Strings now show as expected)

You can also.
In form builder have either blank or any other controls (NOT your custom) added. 
Render
Go back to builder and add your custom control
Strings in your control show as expected as its a second render.

lucasnetau:
Fix #1577 

The language files are loaded async, however they are expected to be loaded by formRender prior to rendering of any controls. This has a possible BC break in that the formRender:render() method is now async and therefore may not have completed if there are post render calls.

### DIFF
--- a/src/js/form-render.js
+++ b/src/js/form-render.js
@@ -64,12 +64,13 @@ class FormRender {
     this.options = jQuery.extend(true, defaults, options)
     this.instanceContainers = []
 
+    this.mi18nLoading = null
+    if (!mi18n.current) {
+      this.mi18nLoading = mi18n.init(this.options.i18n)
+    }
+
     //Override any sanitizer configuration
     setSanitizerConfig(this.options.sanitizerOptions)
-
-    if (!mi18n.current) {
-      mi18n.init(this.options.i18n)
-    }
 
     // parse any passed formData
     if (this.options.formData) {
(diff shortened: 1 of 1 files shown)


src/js/form-render.js
```diff
@@ -197,6 +198,11 @@ class FormRender {
       }
     }
 
+    if (this.mi18nLoading) {
+      await this.mi18nLoading
```
kevinchappell, line 202: :+1:

kevinchappell (approved):

lucasnetau:
Don't merge this until tests are modified

PR #1591, opened 2024-11-14
fix: ensure that other value is required when other option is selected for radio and checkbox groups

lucasnetau:
If Other has been selected we expect that a value will be provided into otherVal. Ensure this by toggling the required attribute on the otherVal text input when the otherOption input is checked


lucasnetau:
We need some tests for these types of issues. Ensuring forms are valid etc. Will look into it.

kevinchappell (approved):
tested locally. good fix!

PR #1609, opened 2025-05-01
Fix/1605 Options unique name not regenerated on clone

### LINKED ISSUE #1605: Copying a select field groups the Options across the fields when it shouldn't
Steps to reproduce:

1. Go to https://formbuilder.online/
2. Add a Select field
3. Copy that Select field
4. Now select any Option in Select field 1, the Option you selected in Select field 2 is cleared. It seems that the Options are grouped across the fields when it shouldn't be.

lucasnetau:
Cleanup for name generation in various places

* Use a unique name for the option group selectors, prevents clash when two fields have the same name
* Regenerate the unique name for the option group selectors on clone Fixes #1605 
* Regenerate item name on clone using the same method as creation (Utils.nameAttr) instead of rolling our own, prevents name clash on multiple clones in same millisecond

### DIFF
--- a/src/js/form-builder.js
+++ b/src/js/form-builder.js
@@ -420,11 +420,11 @@ function FormBuilder(opts, element, $) {
     }
 
     const optionActionsWrap = m('div', optionActions, { className: 'option-actions' })
+    const optionGroupName = nameAttr({type: 'grp-options'}) + '-options'
     const options = m(
       'ol',
-      fieldValues.map((option, index, _, fieldName = fieldData.name) => {
+      fieldValues.map((option, index) => {
         const optionData = config.opts.onAddOption(option, { type, index, isMultiple })
-        const optionGroupName = fieldName + '-options'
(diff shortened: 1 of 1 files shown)


src/js/form-builder.js
```diff
@@ -1609,8 +1609,7 @@ function FormBuilder(opts, element, $) {
 
     const currentId = currentItem.attr('id')
     const type = currentItem.attr('type')
-    const ts = new Date().getTime()
```
kevinchappell, line 1612: in the future there is also the possibility to use `performance.now` in the `nameAttr` function for higher resolution time stamps

kevinchappell (approved):
looks good, works good. :+1:

src/js/form-builder.js
lucasnetau: Yes. We really should be using `Date.now()` instead of `new Date().getTime()` to remove the object creation. We can also replace that call with `performance.timeOrigin + performance.now()`

PR #1611, opened 2025-05-05
Improve formBuilder event dispatch and improve capturing of all stage modifications

### LINKED ISSUE #1596: Sort buttons don't trigger formSaved event
### Description:

This extends #49. I am trying to show a live preview of the form. 
The issue can be spotted into the codepen from the original issue (https://codepen.io/kevinchappell/pen/ZQqwLp).

Due to not having a dedicated event for listening to formBuilder data changes such as "onFormDataChange" event or similar, I resorted to listening to the following events simultaneously "formSaved", "fieldEditOpened", "fieldEditClosed". 

What I want to achieve is to execute a function (rendering the form) when any of the following happens:

1. any field is added
2. any field is removed
3. any field attribute is edited
4. fields order changes

The "formSaved" achieves 1, 2 and partly 4. The event is triggered when the fields order changes from drag&drop with sortable, but it doesn't trigger when sort buttons are clicked (field is moved higher or lower).

### Environment Details:

 - formBuilder Version: 3.19.13
 - Browser: Chrome
 - OS: MacOS 14.6.1

### Expected Behavior
"formSaved" event should be triggered when: 

- fields are moved up or down using the "sort" buttons.
- fields are dragged and dropped

### Actual Behavior
"formSaved" event is **NOT** triggered when fields are moved up or down using the "sort" buttons.

### Steps to Reproduce
https://codepen.io/kevinchappell/pen/ZQqwLp 

1. Add 3+ fields 
2. Drag and drop any field => formRendered form is updated
3. Click any "sort" button on any field => formRendered form is **NOT** updated

lucasnetau:
This PR does two things:
- Call helper.save() after fields are reordered via the manual sorting buttons Fixes #1596
- Refactor events:
  - Remove events.js legacy IE11 support
  - Create events with Bubbling enabled
  - Dispatch events at the stage level (or where appropriate) instead of at document level
- Added save calls when field options were re-ordered or added
- Change to using the input event instead of 'click blur change keyup' to detect changed to attributes in modern browsera
- Ensure that label changes are saved


kevinchappell (approved):
this was long overdue. great work

PR #1636, opened 2026-05-02
fix: #1635 field label not preserved when changing subtype

### LINKED ISSUE #1635: Changing Text Field type resets the label
Hi,

Steps to reproduce

1. Go to https://formbuilder.online/
2. Add a Text Field
3. Change the Label value
4. Change the Type value

The Label is reset to the initial value.

Thank you for your time. <3

kevinchappell:
- **fix: vendor assets missing on first npm start**
- **fix: field label reset when changing the subtype without saving**
- **chore: add test to verify fix**

resolves #1635

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,9 +96,9 @@
     "lint": "eslint ./src || true",
     "lint:fix": "eslint ./src --fix",
     "semantic-release": "semantic-release",
-    "start:devServer": "webpack-dev-server --mode development --config tools/webpack.config",
-    "prestart": "run-p build:vendor copy:lang",
+    "start:devServer": "run-p build:vendor copy:lang && webpack-dev-server --mode development --config tools/webpack.config",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -96,9 +96,9 @@
     "lint": "eslint ./src || true",
     "lint:fix": "eslint ./src --fix",
     "semantic-release": "semantic-release",
-    "start:devServer": "webpack-dev-server --mode development --config tools/webpack.config",
-    "prestart": "run-p build:vendor copy:lang",
```
kevinchappell, line 100: prestart was not firing for me because of `ignore-scripts` setting in .npmrc which was added by me and many others after the [npm Axois supply chain compromise](https://www.microsoft.com/en-us/security/blog/2026/04/01/mitigating-the-axios-npm-supply-chain-compromise/)

